PR #348, opened 2021-05-18
fix: fixes gas cost and provider bugs with hardhat coverage

### LINKED ISSUE #202: fix: `npm run coverage` fails exceeding block gas limit
Looks like `hardhat coverage` significantly increases gas costs with instrumentation. From what I see in the below blog post (linked in hardhat docs), I'm not sure if this is something we can actually fix in our repo. 

Linked in hardhat docs: https://blog.colony.io/code-coverage-for-solidity-eecfa88668c2/
> It is entirely possible that when running SolCover, your tests will fail if they test or rely on properties of your contracts related to how much gas they use. This is because the instrumentation process significantly increases the cost for a deployment (up to ~4 times), as well as the cost of function calls due to the extra events. I have already made the gas cost for EVM’s LOG call zero to help this, but further work should probably be done on this front. Given that this tool already uses its own modified TestRPC , this is easy implement with few other side effects.

More info on gas distortion during coverage here: https://github.com/sc-forks/solidity-coverage/blob/master/docs/faq.md#notes-on-gas-distortion.

ltchang2019:
- something off with using provider wallets as signers in the coverage testing environment
- all actual **signers** must be of type ethers.signer (with `await ethers.getSigners()`) and should NOT be retrieved using `provider.getWallets()`
- `SignersProvider` class hands out all addresses in all places except `GovernanceRouter` tests currently
- will replace wallet usage in GovernanceRouter later on in separate issue or here if #334 is merged before this
Closes #202

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store
--- a/solidity/optics-core/.solcover.js
+++ b/solidity/optics-core/.solcover.js
@@ -1,3 +1,3 @@
 module.exports = {
-  skipFiles: ["test"],
+  skipFiles: ['test'],
 };
--- a/solidity/optics-core/.solcover.js
+++ b/solidity/optics-core/.solcover.js
@@ -1,3 +1,7 @@
 module.exports = {
   skipFiles: ['test'],
+  mocha: {
+    grep: '@skip-on-coverage', // Find everything with this tag
+    invert: true, // Run the grep's inverse set.
+  },
 };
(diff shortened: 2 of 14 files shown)


ltchang2019:
@prestwich I think the only thing missing from this PR now is just getting coverage to show the correct numbers. Any tips on how to fix or did you want to take a look?

prestwich:
Closing as stale. Someday we will get back to this

PR #353, opened 2021-05-20
feat: RecoveryManager role on GovernanceRouter

### LINKED ISSUE #299: feature: Add recoveryManager role to `GovernanceRouter`
- `recoveryManager` can initiate the recovery timelock 
- when timelock ends, anyone can trigger "recovery" state
- during "recovery" state,`recoveryManager`'s role is "activated" (it can start doing things)
- during recovery, `recoveryManager` should only be able to take actions on local chain (not sending messages to remote chains, which governor can do) 
- `recoveryManager` should be able to `call` contracts on the local chain 
- perhaps add a flag indicating when the `GovernanceRouter` is under "recovery" ? change the modifiers so the full governor powers are only available when not under recovery? (need to think more about this)

more documentation [here](https://www.notion.so/clabsco/recoveryManager-69a2ab0e6da048039f17bf66ecaa30a6)

anna-carroll:
closes #299


solidity/optics-core/contracts/governance/GovernanceRouter.sol
```diff
@@ -205,11 +271,44 @@ contract GovernanceRouter is Initializable, IMessageRecipient {
...
+     * @dev callable by the recovery manager to end recovery mode
+     */
+    function exitRecovery() external onlyInRecovery onlyRecoveryManager {
+        delete recoveryActiveAt;
+
+        emit ExitRecovery(recoveryManager);
+    }
+
+    function inRecovery() public view returns (bool) {
+        uint256 _recoveryActiveAt = recoveryActiveAt;
```
ltchang2019, line 306: nit: Is this load into `_recoveryActiveAt` necessary?

solidity/optics-core/contracts/governance/GovernanceRouter.sol
```diff
@@ -205,11 +271,44 @@ contract GovernanceRouter is Initializable, IMessageRecipient {
...
+     */
+    function initiateRecoveryTimelock()
+        external
+        onlyNotInRecovery
+        onlyRecoveryManager
+    {
+        require(recoveryActiveAt == 0, "timelock already initiated");
+
+        recoveryActiveAt = block.number + recoveryTimelock;
+
```
ltchang2019, line 291: Is this overflow-safe? Looks like we condition on `recoveryActiveAt` and `recoveryTimelock` a few times so just wanted to check

solidity/optics-core/contracts/governance/GovernanceRouter.sol
```diff
@@ -205,11 +271,44 @@ contract GovernanceRouter is Initializable, IMessageRecipient {
...
+        delete recoveryActiveAt;
+
+        emit ExitRecovery(recoveryManager);
+    }
+
+    function inRecovery() public view returns (bool) {
+        uint256 _recoveryActiveAt = recoveryActiveAt;
+        bool _recoveryInitiated = _recoveryActiveAt != 0;
+        bool _recoveryTimelockExpired = _recoveryActiveAt <= block.number;
+        return _recoveryInitiated && _recoveryTimelockExpired;
```
ErinHales, line 309: _recoveryTimelockExpired is actually checking if it has _not_ expired.  Would rename or change to:
```
bool _recoveryTimelockExpired = _recoveryActiveAt > block.number;
return _recoveryInitiated && !_recoveryTimelockExpired;
```

solidity/optics-core/contracts/governance/GovernanceRouter.sol
```diff
@@ -15,37 +15,57 @@ contract GovernanceRouter is Initializable, IMessageRecipient {
     using TypedMemView for bytes29;
     using GovernanceMessage for bytes29;
 
-    XAppConnectionManager public xAppConnectionManager;
-
     uint32 public immutable localDomain;
-    uint32 public governorDomain; // domain of Governor chain -- for accepting incoming messages from Governor
+    uint256 public immutable recoveryTimelock; // number of blocks before recovery can be activated
+
+    uint256 public recoveryActiveAt; // block number when recovery timelock expires; 0 if timelock has not been initiated
```
ErinHales, line 21: consider renaming, sounds like recovery starts at that block.  `recoveryExpiresAt`?

ErinHales (commented):
nice! 😊

solidity/optics-core/contracts/governance/GovernanceRouter.sol
anna-carroll: it's really just to save the 800 gas of an extra SLOAD. since this function will be called in a modifier that will be added to nearly every external function

solidity/optics-core/contracts/governance/GovernanceRouter.sol
anna-carroll: 🤔 it shouldn't overflow given the real-world expected values of these vars...

solidity/optics-core/contracts/governance/GovernanceRouter.sol
anna-carroll: `recovery` is the state in which the RecoveryManager is able to act in place of the governor; the `recovery` **timelock** expires at this block, while `recovery` itself is active at this block

solidity/optics-core/contracts/governance/GovernanceRouter.sol
anna-carroll: so, no, it's not necessary - it's a very small optimization

solidity/optics-core/contracts/governance/GovernanceRouter.sol
anna-carroll: heck, yeah, great catch https://github.com/celo-org/optics-monorepo/pull/353/commits/a5ba7acb4bc86325a3b56f5e71e6de99e54dae38

solidity/optics-core/contracts/governance/GovernanceRouter.sol
anna-carroll: renamed for a bit more clarity / consistency hopefully? https://github.com/celo-org/optics-monorepo/pull/353/commits/a5ba7acb4bc86325a3b56f5e71e6de99e54dae38

anna-carroll: recovery timelock expired = recovery is active

solidity/optics-core/contracts/governance/GovernanceRouter.sol
anna-carroll: hmm i see where the confusion is arising here. the name `recoveryTimelockExpired` is the same as the name `recoveryActive` - but perhaps the latter is more clear. 

let's say `recoveryTimelock` is 25 blocks, and the `recoveryManager` calls `initiateRecovery` at `block.number` 50. this means that `recoveryActiveAt` = 75 (the current block 50, plus 25 block timelock)

the 25-block recovery timelock expires at block 75. recovery is active at block 75. for any block that is greater than or equal to 75, recovery is active.

anna-carroll: i renamed to `recoveryActive` to make this hopefully more clear?

solidity/optics-core/contracts/governance/GovernanceRouter.sol
prestwich: why did we choose to do a blocknumber instead of a timestamp?

solidity/optics-core/contracts/governance/GovernanceRouter.sol
```diff
@@ -191,9 +259,9 @@ contract GovernanceRouter is Initializable, IMessageRecipient {
      * @param _domain The domain
      * @param _router The new router
      */
-    function setRouterDuringSetup(uint32 _domain, bytes32 _router)
+    function setRouterLocal(uint32 _domain, bytes32 _router)
```
prestwich, line 262: could be `enrollRouter` ?

solidity/optics-core/contracts/governance/GovernanceRouter.sol
prestwich: if you leave it as is, document that setting `recoveryTimelock` near u256 max has the opposite of intended behavior

solidity/optics-core/contracts/governance/GovernanceRouter.sol
prestwich: the savings from doing this went down in eip2929, but are still important

solidity/optics-core/js/types.js
```diff
@@ -2,6 +2,8 @@
  * ChainConfig {
  *      domain: int,
  *      updater: address,
+ *      recoveryTimelock: int,
+ *      recoveryManager: address,  // NOTE: this may change if we add a multisig to the deploy setup
```
prestwich, line 6: need to add these to the TS deploy process

prestwich (changes requested):
see comments

solidity/optics-core/contracts/governance/GovernanceRouter.sol
anna-carroll: switched to `SafeMath` : https://github.com/celo-org/optics-monorepo/pull/353/commits/d14e978939307ac81f261f2d88cf2b89d50ae0c0

solidity/optics-core/contracts/governance/GovernanceRouter.sol
anna-carroll: switched https://github.com/celo-org/optics-monorepo/pull/353/commits/d14e978939307ac81f261f2d88cf2b89d50ae0c0

solidity/optics-core/contracts/governance/GovernanceRouter.sol
anna-carroll: this function is used to add routers (enroll), remove routers (un-enroll), and change router addresses

solidity/optics-core/js/types.js
anna-carroll: https://github.com/celo-org/optics-monorepo/pull/353/commits/1886f01a7560ece708985ca15b113b4e3ad59b8d

I can send you the updated `tmp.ts` for this and `test_deploy.env`

solidity/optics-core/contracts/governance/GovernanceRouter.sol
```diff
@@ -55,23 +77,50 @@ contract GovernanceRouter is Initializable, IMessageRecipient {
...
+        _;
+    }
+
+    modifier onlyNotInRecovery() {
+        require(!inRecovery(), "in recovery");
+        _;
+    }
+
+    modifier onlyGovernorOrRecoveryManager() {
+        if (inRecovery()) {
```
prestwich, line 116: nit: whichever check is more likely to pass should be first. In this case, the governor check

prestwich (changes requested):
1 nit

solidity/optics-core/contracts/governance/GovernanceRouter.sol
anna-carroll: https://github.com/celo-org/optics-monorepo/pull/353/commits/42c01a074bed74d849940cca361bbdc7d5f1dd02

prestwich (approved):

PR #356, opened 2021-05-20
[Feature] initial dockerfile implementation

yourbuddyconner:
Implements a Dockerfile that produces an ubuntu-based Docker image containing the agent binaries. 

Notes: 
- Moved `config/` to `rust/config`
- Moved `abis/` to `rust/optics-ethereum/abis`
- Updated scripts, removed redundant pre-commit step that checks `abis/` for updates


prestwich:
unless you update the scripts to copy ABIs to the new location before rust compilation, this will break next time the ABI of the contracts changes

prestwich:
now that typescript deploy is in `main` you'll need to update its config output path as well

pre-commit.sh
```diff
@@ -150,14 +150,6 @@ else
     echo "+Skipping optics-xapps tests"
 fi
 
-# Git add abis if updated
```
prestwich, line 153: reroute instead of delete?

rust/optics-base/src/macros.rs
```diff
@@ -105,8 +105,8 @@ macro_rules! decl_settings {
 
                 let env = std::env::var("RUN_ENV").unwrap_or_else(|_| "default".into());
 
-                s.merge(config::File::with_name(&format!("../config/{}/base", env)))?;
-                s.merge(config::File::with_name(&format!("../config/{}/{}-partial", env, $name)).required(false))?;
+                s.merge(config::File::with_name(&format!("./config/{}/base", env)))?;
+                s.merge(config::File::with_name(&format!("./config/{}/{}-partial", env, $name)).required(false))?;
```
prestwich, line 109: need to rebase onto `main` and update with new `BASE_FILE` env

pre-commit.sh
```diff
@@ -152,8 +152,8 @@ fi
 
 # Git add abis if updated
 if ! git diff-index --quiet HEAD -- ./abis; then
```
prestwich, line 154: need to modify this check to check the new location

prestwich (approved):

prestwich:
manually squashed and merged

PR #362, opened 2021-05-24
fix: handles undefined GITHUB_TOKEN

ErinHales:
Only contributors have access to repository secrets.  There isn't a great work around to this so I've disabled the workflow fun if GITHUB_TOKEN is not defined


anna-carroll (approved):
🦒

PR #382, opened 2021-06-08
refactor: combine lib and script folders to js folder

ErinHales:


ErinHales:
Was seeing this error when running the pre-commit script, but not when I ran the `test-all` script.  Not sure if this is related to something I did. 
Screen Shot 2021-06-08 at 11 17 32 AM

prestwich:
this might be related to the way we're mixing providers? the tests could really use an overhaul in TS 😬

ErinHales:
Have confirmed this is happening on main as well, so I will open an issue for it.  I can open an issue for ts tests as well

prestwich:
to resolve conflicts, rebase onto main and run `npx hardhat typechain` from the optics-core directory

ErinHales:
Oh, I didn't realize there were merge conflicts!  Ok, those are taken care of

prestwich (approved):

prestwich:
more merge conflicts in autogenerated files 🙌

prestwich (approved):

PR #392, opened 2021-06-11
chore: add typechain generation to standard pre-commit script

### LINKED ISSUE #374: add typechain generation to standard pre-commit script
currently you have to manually run `npx hardhat typechain`. Add this as a step in the `npm build` command

YoduYodu:
closes #374

### DIFF
--- a/pre-commit.sh
+++ b/pre-commit.sh
@@ -28,6 +28,7 @@ echo "Making pre-run checks"
 ensure npm
 ensure cargo
 ensure jq
+ensure npx
 
 # TODO: DRY these
 if (echo min version $MIN_GIT; git --version) | sort -Vk3 | tail -1 | grep -q git
@@ -67,28 +68,36 @@ trap 'abort' 0
 set -e
 git update-index -q --refresh
 
-# Conditionally compile and update optics-core abis
+# Conditionally compile, update optics-core abis and generate typechain
 if ! git diff-index --quiet HEAD -- ./solidity/optics-core; then
-    echo "+Updating core ABIs"
+    echo "+Updating core ABIs, generating typechain"
     cd ./solidity/optics-core
     npm run compile
-    # add files in case linter modified them
-    git add .
+    # add abis, typechain
     cd ../..
+    git add rust/optics-ethereum/abis
+    git add typescript/src/typechain
+    # add linter modified files
+    git add solidity/optics-core/contracts
+    git add solidity/optics-core/libs
+    git add solidity/optics-core/interfaces
 else
-    echo "+Skipping core ABI updates"
+    echo "+Skipping core ABI updates and typechain generation"
 fi
 
-# Conditionally compile and update optics-xapps abis
+# Conditionally compile, update optics-xapps abis and generate typechain
 if ! git diff-index --quiet HEAD -- ./solidity/optics-xapps; then
-    echo "+Updating xapps ABIs"
+    echo "+Updating xapps ABIs, generating typechain"
     cd ./solidity/optics-xapps
     npm run compile
-    # add files in case linter modified them
-    git add .
+    # add typechain
     cd ../..
+    git add typescript/src/typechain
+    # add linter modified files
+    git add solidity/optics-xapps/contracts
+    git add solidity/optics-xapps/interfaces
 else
-    echo "+Skipping xapps ABI updates"
+    echo "+Skipping xapps ABI updates and typechain generation"
 fi
 
 # Conditionally run Rust bins to output into vector JSON files
--- a/solidity/optics-core/package.json
+++ b/solidity/optics-core/package.json
@@ -27,7 +27,7 @@
   },
   "scripts": {
     "prettier": "prettier --write ./contracts ./test ./js",
-    "compile": "hardhat compile && npm run prettier && ./js/update_abis.sh",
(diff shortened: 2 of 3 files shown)


pre-commit.sh
```diff
@@ -67,28 +67,32 @@ trap 'abort' 0
...
 
-# Conditionally compile and update optics-core abis
+# Conditionally compile and update optics-core abis, generate typechain
 if ! git diff-index --quiet HEAD -- ./solidity/optics-core; then
     echo "+Updating core ABIs"
     cd ./solidity/optics-core
     npm run compile
-    # add files in case linter modified them
+    echo "+Generating core typechain"
+    npx hardhat typechain
```
prestwich, line 76: Add a check at the beginning to see if npx is installed

prestwich (changes requested):

pre-commit.sh
```diff
@@ -67,28 +67,32 @@ trap 'abort' 0
...
+    # add files in case linter modified them, or typechain generation
     git add .
```
prestwich, line 78: this `git add` will not add the typechain or abis anymore. this is a bug introduced by moving things in the repo.

change to 
```
cd ../..
git add rust/optics-ethereum/abis
git add typescript/src/typechain
```

pre-commit.sh
```diff
@@ -67,28 +67,32 @@ trap 'abort' 0
...
+# Conditionally compile and update optics-xapps abis, generate typechain
 if ! git diff-index --quiet HEAD -- ./solidity/optics-xapps; then
     echo "+Updating xapps ABIs"
     cd ./solidity/optics-xapps
     npm run compile
-    # add files in case linter modified them
+    echo "+Generating xapps typechain"
+    npx hardhat typechain
+    # add files in case linter modified them, or typechain generation
     git add .
```
prestwich, line 92: same here. change to 

```
cd ../..
git add typescript/src/typechain
```

prestwich (changes requested):

pre-commit.sh
ErinHales: oops, good catch

pre-commit.sh
ErinHales, line 76: I think it would make sense to add this command to the compile script

pre-commit.sh
```diff
@@ -67,28 +68,29 @@ trap 'abort' 0
...
     npm run compile
-    # add files in case linter modified them
-    git add .
+    # add updated files
     cd ../..
+    git add typescript/src/typechain
 else
-    echo "+Skipping xapps ABI updates"
+    echo "+Skipping xapps ABI updates and typechain generation"
 fi
```
YoduYodu, line 94: Regarding comments, does it make sense to just say "compile" rather than "compile, update abis, typechain...", let the script document itself?

pre-commit.sh
```diff
@@ -67,30 +68,27 @@ trap 'abort' 0
...
     cd ./solidity/optics-core
     npm run compile
-    echo "+Generating core typechain"
-    npx hardhat typechain
-    # add files in case linter modified them, or typechain generation
-    git add .
+    # add updated files
     cd ../..
+    git add rust/optics-ethereum/abis
+    git add typescript/src/typechain
```
ErinHales, line 79: This won't catch files that were modified by the linter.  Should we just `git add .` here? @prestwich

prestwich: I think doing `git add .` would result in accidentally committing a bunch of files. instead can we explicitly add specific directories? in this case `solidity/optics-core/contracts` and `solidity/optics-core/libs` and `solidity/optics-core/interfaces`

solidity/optics-core/package.json
```diff
@@ -27,7 +27,7 @@
   },
   "scripts": {
     "prettier": "prettier --write ./contracts ./test ./scripts ./lib",
-    "compile": "hardhat compile && npm run prettier && ./scripts/update_abis.sh",
+    "compile": "hardhat compile && npm run prettier && ./scripts/update_abis.sh && npx hardhat typechain",
```
prestwich, line 30: no need for the npx here. the scripts already run in the node context

pre-commit.sh
prestwich: fine either way

solidity/optics-xapps/package.json
```diff
@@ -27,7 +27,7 @@
   },
   "scripts": {
     "prettier": "prettier --write './contracts' ./test",
-    "compile": "hardhat compile && npm run prettier",
+    "compile": "hardhat compile && npm run prettier && npx hardhat typechain",
```
prestwich, line 30: ditto

pre-commit.sh
```diff
@@ -67,28 +68,37 @@ trap 'abort' 0
...
     cd ./solidity/optics-xapps
     npm run compile
-    # add files in case linter modified them
-    git add .
+    # add typechain
     cd ../..
+    git add typescript/src/typechain
+    # add linter modified files
+    git add solidity/optics-xapps/contracts
+    git add solidity/optics-xapps/libs
```
ErinHales, line 98: There isn't a libs folder in `optics-xapps` (:

ErinHales (approved):

prestwich (dismissed):
LGTM

ErinHales (approved):

PR #395, opened 2021-06-14
test: setup for ts tests

ErinHales:
- migrate HardhatRuntimeEnvironment
 - add deploy for test contracts
 - test Common


prestwich (changes requested):
okay, some good work here. I'd like to see us split up the tests and the dev deploys to a new typescript package. `typescript/optics-deploy` and `typescript/optics-test`

ErinHales:
> okay, some good work here. I'd like to see us split up the tests and the dev deploys to a new typescript package. `typescript/optics-deploy` and `typescript/optics-test`

Yep, that definitely makes sense.  I refactored the typescript directory a bit.  I didn't rename the `test` directory to `optics-test` though, because hardhat is stubborn

prestwich:
do you think it makes sense to have these be entirely separate packages? each with a package.json?

Then only the optics-test package would need hardhat

ErinHales:
> do you think it makes sense to have these be entirely separate packages? each with a package.json?
> 
> Then only the optics-test package would need hardhat

Possibly? Do you think we'd want to make optics-deploy into an npm package?

typescript/test/lib/index.ts
```diff
@@ -0,0 +1,283 @@
...
+import * as deployHelpers from '../../optics-deploy/deployOptics';
+import { getHexStringByteLength } from './utils';
+<<<<<<< Updated upstream
+import { TestHome__factory, TestReplica__factory } from '../../typechain/optics-core';
+=======
+import {
+  TestHome__factory,
+  TestReplica__factory,
+} from '../../typechain/optics-core';
+>>>>>>> Stashed changes
```
YoduYodu, line 16: ☝️

ErinHales:
TODO:Install packages in `typescript/typechain`

ErinHales:
Closing in favor of #400

PR #397, opened 2021-06-15
refactor: home 0-indexes sequences

### LINKED ISSUE #396: domain sequences in home are 1-indexed
lmao those should obv be 0-indexed

prestwich:
@ErinHales can you fix the last test on this? The cross-chain governance proof needs to be regenerated. cc @anna-carroll if you get stuck

closes #396

### DIFF
--- a/solidity/optics-core/contracts/Home.sol
+++ b/solidity/optics-core/contracts/Home.sol
@@ -128,7 +128,6 @@ contract Home is Initializable, MerkleTreeManager, QueueManager, Common {
     ) external notFailed {
         require(_body.length <= MAX_MESSAGE_BODY_BYTES, "!too big");
         uint32 _sequence = sequences[_destination];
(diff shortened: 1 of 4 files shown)


solidity/optics-core/contracts/Home.sol
```diff
@@ -151,6 +150,8 @@ contract Home is Initializable, MerkleTreeManager, QueueManager, Common {
             _leaf,
             _message
         );
+
+        sequences[_destination] = _sequence + 1;
```
ErinHales, line 154: Note: moved this to the bottom of the function, so sequence wouldn't be bumped if another error occurred during enqueue

prestwich: any error would result in total reversion of the call regardless. still it's good hygiene to have this at the end anyway

ErinHales (approved):

PR #401, opened 2021-06-16
test: home ts

ErinHales:


typescript/optics-tests/test/home.test.ts
```diff
@@ -1,9 +1,258 @@
...
+    const newRoot = await enqueueMessageAndGetRoot('message');
+    const { signature: fakeSignature } = await fakeUpdater.signUpdate(
+      currentRoot,
+      newRoot,
+    );
+    await expect(
+      home.proxy.update(currentRoot, newRoot, fakeSignature),
+    ).to.be.revertedWith('bad sig');
+  });
+  it('Fails on valid double update proof', async () => {
```
prestwich, line 229: leave a blank line between the tests

prestwich (changes requested):

typescript/optics-tests/test/home.test.ts
```diff
@@ -1,9 +1,268 @@
...
+import { SignerWithAddress } from '@nomiclabs/hardhat-ethers/signers';
 const { expect } = require('chai');
-const { optics } = require('hardhat');
+
+import * as chain from '../../optics-deploy/src/chain';
+import * as deploys from '../../optics-deploy/src/deployOptics';
+import { getTestDeploy } from './testChain';
+import { OpticsState, Updater } from '../lib';
+
+const UpdaterManager = require('../../../solidity/optics-core/artifacts/contracts/UpdaterManager.sol/UpdaterManager.json');
```
prestwich, line 11: can we change these to not be requires so we're not mixing syntax. you can import json in TS by adding ` "resolveJsonModule": true,` to the compiler options in the `.tsconfig`

prestwich (approved):

prestwich (commented):
looking good, still need to change requires to imports

ErinHales:
> looking good, still need to change requires to imports

I thought I got all of them...  Looks like I missed hardhat.config.ts, but I don't think there are any others?

PR #404, opened 2021-06-17
Remove return values from `handle`

### LINKED ISSUE #403: remove return value from `handle` interface
make a cleaner solution to the hacky fix to Solidity's try/catch issue, as [discussed in Slack](https://celo-org.slack.com/archives/C01KT8JKQ67/p1623795423087300)

anna-carroll:
Closes #403


solidity/optics-core/contracts/Replica.sol
```diff
@@ -222,16 +219,18 @@ contract Replica is Initializable, Common, QueueManager {
...
-        (_success, _result) = _recipient.call{gas: PROCESS_GAS}(_calldata);
+        try
+            IMessageRecipient(_recipient).handle{gas: PROCESS_GAS}(
+                _m.origin(),
+                _m.sender(),
+                _m.body().clone()
+            )
+        {
+            _success = true;
+        } catch (bytes memory _err) {
```
prestwich, line 230: I think this has the same problem. if the MessageRecipient reverts with data that can't be abi decoded as a `bytes`, it will cause our contract to revert

prestwich (changes requested):

solidity/optics-core/contracts/Replica.sol
prestwich: for an example

```
assembly {
    revert(0,0)    
}
```

this would revert and the returndata would be set to the empty byte array`[]`, which would then fail when solidity attempts to decode as `bytes memory _err`

anna-carroll: Hmm interesting, so this actually worked in the test that I wrote: https://github.com/celo-org/optics-monorepo/pull/404/commits/3dc4dbd01f96876f3d1ded292670b2c5b345ab06

anna-carroll: ah yeah, so i got my info about try/catch [here](https://blog.ethereum.org/2020/01/29/solidity-0.6-try-catch/) and it specifically says 

> The advantage of the second method is that it is also executed if ABI decoding the error string fails or if no reason was provided.

which i checked before writing this and then promptly forgot again

solidity/optics-core/test/Replica.test.js
```diff
@@ -352,6 +344,33 @@ describe('Replica', async () => {
     await expect(replica.process(opticsMessage)).to.be.revertedWith('!pending');
   });
 
+  it('Processes a message that reverts during handle', async () => {
```
prestwich, line 347: can we add some similar tests for:
```
assembly {
    return(0,0)
}

assembly {
    mstore(0, 0xabcdefg)
    return(0,32)
}

assembly {
    mstore(0, 0xabcdefg)
    revert(0, 32)
}
```

prestwich (changes requested):

solidity/optics-core/test/Replica.test.js
anna-carroll: 🤠 https://github.com/celo-org/optics-monorepo/pull/404/commits/c6d29c6974a1eaafab7ea1eabd39a8675be415bf

anna-carroll: I removed the expectation that `success` would be `false`, as two of these "Bad" recipients don't actually revert

prestwich (approved):
👍

PR #409, opened 2021-06-20
Processing Refactor

### LINKED ISSUE #405: feature: Replica should allow proofs under any past root
This simplifies processor design by ensuring that a proof under a valid update stays valid indefinitely

prestwich:
Starts a refactor of message processing

1. closes #405 by introducing an `acceptableRoot` function to the replica
2. refactors the processor to cache proofs for each leaf under an acceptable root


prestwich:
@anna-carroll on monday, can you review this directionally?

solidity/optics-core/contracts/Replica.sol
```diff
@@ -49,6 +47,14 @@ contract Replica is Initializable, Common, QueueManager {
 
     constructor(uint32 _localDomain) Common(_localDomain) {} // solhint-disable-line no-empty-blocks
 
+    function acceptableRoot(bytes32 _root) internal view returns (bool) {
+        uint256 _time = confirmAt[_root];
+        if (_time == 0) {
+            return false;
+        }
+        return block.timestamp >= _time || _root == current;
```
anna-carroll, line 55: to make sure I understand - we should never hit the `_root == current` condition within the `confirm()` function because `current` root should never be in the `queue`, right?

the `_root == current` condition is relevant for the `prove` function, so that we can prove the current root OR any other root whose timelock has passed

anna-carroll: thinking out loud - this means that we can `prove` roots that haven't been confirmed yet - e.g. they may still be in the queue. could this cause any unintended behavior?

anna-carroll: i am trying to think through whether this kind of obviates the need for the queue / `confirm()` function - this way, any root that times out automatically becomes provable / processable. no need to confirm.

i am also trying to think of whether this might break some of our assumptions about messages being processed in order.... i think it does

anna-carroll (commented):
looking directionally good 👍

solidity/optics-core/contracts/Replica.sol
prestwich:
```
thinking out loud - this means that we can prove roots that haven't been confirmed yet - e.g. they may still be in the queue. could this cause any unintended behavior?
```

I think this is fine, as we guarantee that they _can_ be confirmed. I.e. the next call to confirm will certainly confirm them. i.e. they are _already_ good, they just haven't been rubber stamped yet

```
i am trying to think through whether this kind of obviates the need for the queue / confirm() function - this way, any root that times out automatically becomes provable / processable. no need to confirm.
```

I had the same thought, but decided that it's too big of a refactor. there's relatively little downside to keeping the queue, and it simplifies finding the latest root

```
i am also trying to think of whether this might break some of our assumptions about messages being processed in order.... i think it does
```

No, messages will still be processed in order unless we also remove the sequence no. checks

asaj: > I had the same thought, but decided that it's too big of a refactor. there's relatively little downside to keeping the queue, and it simplifies finding the latest root

Trying to understand the system better, what else would need to change in order to drop `confirm()`?

IIUC, at least on the solidity side, it seems pretty straightforward. `current` is no longer needed in `Replica`, and `Replica.initialize()` would need to allow the current root to be added to the queue, and I think that's it?

I'm guessing the main complication with removing `confirm` would be that `current` is used by the processor to decide when to prove and process messages, and that changing the processor to use block time instead would be more trouble than it's worth. Is that right?

solidity/optics-core/contracts/Replica.sol
```diff
@@ -63,6 +69,7 @@ contract Replica is Initializable, Common, QueueManager {
         queue.initialize();
 
         current = _current;
+        confirmAt[_current] = 1;
```
asaj, line 72: nit: Consider a comment explaining why this is needed

solidity/optics-core/contracts/Replica.sol
asaj: > I had the same thought, but decided that it's too big of a refactor. there's relatively little downside to keeping the queue, and it simplifies finding the latest root

Trying to understand the system better, what else would need to change in order to drop confirm()?

IIUC, at least on the solidity side, it seems pretty straightforward. current is no longer needed in Replica, and Replica.initialize() would need to allow the current root to be added to the queue, and I think that's it?

I'm guessing the main complication with removing confirm would be that current is used by the processor to decide when to prove and process messages, and that changing the processor to use block time instead would be more trouble than it's worth. Is that right?

asaj: I suppose actually you would cut the `queue` entirely (in addition to `current`) and instead only keep track of the `latest` root and confirmation times

prestwich: yeah, you're correct. the gains are not worth the refactor. calling `confirm` is a pretty small part of the costs of the system and changing it woudl be a fair number of dev hours at this point

solidity/optics-core/contracts/Replica.sol
```diff
@@ -51,6 +49,14 @@ contract Replica is Initializable, Common, QueueManager {
 
     constructor(uint32 _localDomain) Common(_localDomain) {} // solhint-disable-line no-empty-blocks
 
+    function acceptableRoot(bytes32 _root) internal view returns (bool) {
+        uint256 _time = confirmAt[_root];
+        if (_time == 0) {
+            return false;
+        }
```
anna-carroll, line 56: should we remove the `|| _root == current` and just have 

```
return block.timestamp >= _time;
```
?

anna-carroll, line 56: should we remove the `|| _root == current` and just have
```
return block.timestamp >= _time;
```
?

anna-carroll: I cannot think of a case where `block.timestamp < _time && _root == current`

anna-carroll: e.g. i can't think of a time when the first statement in the `||` is false and the second is true, so we can just simplify to the first statement

anna-carroll (approved):
awesome possum

PR #414, opened 2021-06-23
refactor: several rust improvements

### LINKED ISSUE #412: Stabilize Processor

prestwich:
- add partials to latest deploy
- fix the prefix iterator in db
- run the core updater loop in the current span
- correct a processor bug that would prevent startup after any proofs were cached
- correct a processor bug that would prevent startup with the initial tree

closes #412


prestwich:
@anna-carroll due to the change to the replica interface, this will now require a fresh contract deployment

prestwich:
exercising executive privilege to merge this immediately

anna-carroll (commented):
LGTM

PR #416, opened 2021-06-25
Polish Bridge Code

anna-carroll:


solidity/optics-xapps/contracts/bridge/TokenRegistry.sol
```diff
@@ -94,12 +99,12 @@ abstract contract TokenRegistry is XAppConnectionClient {
         typeAssert(_tokenId, BridgeMessage.Types.TokenId)
         returns (address _token)
     {
-        bytes32 _idHash = _tokenId.keccak();
-        _token = _createClone(tokenTemplate);
-
+        // Deploy the token contract by cloning tokenTemplate
+        _token = _cloneTokenContract();
         // Initial details are set to a hash of the ID
+        bytes32 _idHash = _tokenId.keccak();
```
prestwich, line 105: on second thought, this could cause problems for wallet software that assumes name to be a valid string 🤔 

worth documenting (elsewhere) that we limit the name to 32 bytes in order to block very long name/symbol combos

prestwich: instead of doing the hash of the token id, could we do something like the domain as a string + the first few bytes of the address? 🤔 

see my hex encoder over here:
https://github.com/summa-tx/memview-sol/blob/main/contracts/TypedMemView.sol#L77-L108

anna-carroll: > on second thought, this could cause problems for wallet software that assumes name to be a valid string 🤔

yeah, i definitely don't love this - i have been trying to think of a world where the name/symbol/decimals get sent in the same message as the tokens, whenever tokens are sent to a chain for the first time. but that gets messy quickly, as we discussed.

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
```diff
@@ -40,88 +43,106 @@ contract BridgeRouter is Router, TokenRegistry {
...
+
     /**
-     * @notice Sends a Details message.
+     * @notice Update the token metadata on another chain
      * @param _token The token address
      * @param _destination The destination domain
      */
+    // TODO: people can call this for nonsense non-ERC-20 tokens
+    // name, symbol, decimals could be nonsense
+    // remote chains will deploy a token contract based on this message
```
prestwich, line 105: oh yeah, let's not have the remote chain deploy a token based on this. Instead it should just do nothing if there's no token already deployed for that identifier 🤔

solidity/optics-xapps/contracts/bridge/TokenRegistry.sol
anna-carroll: i think we should have the token name be a string like "`optics.[domain].[tokenAddress]`" so that off-chain apps can recognize when a token deployed by optics doesn't have its name updated yet, and they can query the name on the original token contract on the originating chain using the (domain, tokenAddress) identifier

prestwich: in order to do that we'd need to write 

```
function encodeTokenId(TokenId _id) internal pure returns (string memory)
```

To encode the integer in decimal, 

```
ASCII_0 = 0x30

bytes memory s;  // with length 46 + 
while (n != 0) {
    nextChar = ASCII_0 + n % 10;
    s[i] = nextChar;
    i += 1;
}
```

Then do the hex like my example in typed memview

prestwich: note that we output from right to left for this part

prestwich: python implementation to build intuition
```
In [1]: def int_to_str(i: int) -> str:
    ...:     ASCII_0 = 0x30
    ...:     b = []
    ...:     while (i != 0):
    ...:         b.insert(0, ASCII_0 + i % 10)
    ...:         i //= 10
    ...:     return bytearray(b).decode('utf8')

In [2]: int_to_str(1003811238471203840)
Out[2]: '1003811238471203840'

```

prestwich:
Please open issues for the 2 above comments. then rebase onto main and rebuild the typechain and we'll get this merged

prestwich (approved):

PR #423, opened 2021-06-29
test: add message test

YoduYodu:


typescript/optics-tests/test/message.test.ts
```diff
@@ -0,0 +1,92 @@
+import { ethers } from 'ethers';
+
+const { waffle, optics } = require('hardhat');
```
prestwich, line 3: prefer to decirculate the direct usage of waffle. instead

```
import {ethers, optics} from 'hardhat';
...
    let [signer] = await ethers.getSigners();
```

typescript/optics-tests/test/message.test.ts
```diff
@@ -0,0 +1,92 @@
...
+const { provider } = waffle;
+const { expect } = require('chai');
```
prestwich, line 5: change these to imports instead of requires

ErinHales (approved):

PR #439, opened 2021-07-01
fix: Use low-level call instead of contract call

anna-carroll:


prestwich (approved):
Please open a second PR to adjust `ProcessError` to be as follows:

`event ProcessError(uint32 indexed sequence, address indexed recipient, bytes returnData)` 

makes it easier to us to search for errors on specific requests

PR #444, opened 2021-07-01
Fresh deploy

anna-carroll:


prestwich (changes requested):
your partials are out of date. integers should be strings in the new config format

prestwich (approved):

PR #447, opened 2021-07-01
Copy partial configs during deploy script

### LINKED ISSUE #445: add partial configs during deploy script
add partial configs from `default` folder into each new deploy as part of the deploy script

anna-carroll:
Closes #445


prestwich (approved):
LGTM

PR #448, opened 2021-07-02
Verify contract code after deploy script

anna-carroll:


solidity/optics-core/hardhat.config.js
```diff
@@ -43,4 +46,7 @@ module.exports = {
   mocha: {
     bail: true,
   },
+  etherscan: {
+    apiKey: "[INSERT KEY]"
```
prestwich, line 50: load from env var

prestwich:
is this PR for informational purposes, or is it something we expect to merge eventually?

anna-carroll:
> is this PR for informational purposes, or is it something we expect to merge eventually?

it was originally informational, so I could push something we would both look at, but I have co-opted the branch / PR for the actual feature

anna-carroll:
**TODO:** 
- [ ] load etherscan API key from process.env 
- [ ] switch networks automatically using hardhat-ethers rather than .env variables 
- [ ] create an npm script to first run the deploy script then run the verify script 
- [ ] test the end-to-end flow

solidity/optics-core/hardhat.config.js
anna-carroll: https://github.com/celo-org/optics-monorepo/pull/448/commits/e02e0c71eb052cd3910b14214819799275902d1b

solidity/optics-core/package.json
```diff
@@ -29,16 +29,18 @@
     "prettier": "prettier --write ./contracts ./test ./js",
     "compile": "hardhat compile && hardhat typechain && npm run prettier && ./js/update_abis.sh",
     "coverage": "npm run compile && hardhat coverage",
-    "test": "npm run compile && hardhat test",
+    "test": "npm run compile && hardhat --network hardhat test",
```
prestwich, line 32: still needed?

typescript/src/proxyUtils.ts
```diff
@@ -21,6 +21,7 @@ export type ProxyAddresses = {
  * @param T - The contract
  */
 export async function deployProxy<T extends ethers.Contract>(
+  name: string,
```
prestwich, line 24: you can get the class name using `implementation.constructor.name` instead of passing it as a string

typescript/src/proxyUtils.ts
```diff
@@ -52,13 +74,22 @@ export async function deployProxy<T extends ethers.Contract>(
  * @param T - The contract
  */
 export async function duplicate<T extends ethers.Contract>(
+    name: string,
```
prestwich, line 77: you can get the class name using `prev.implementation.constructor.name` instead of passing it as a string

prestwich (changes requested):

typescript/src/proxyUtils.ts
prestwich: this may not actually work correctly 🤔

anna-carroll: https://github.com/celo-org/optics-monorepo/pull/448/commits/2f7343bf7058c5d05ee8990a183edfc5e3893509

typescript/src/proxyUtils.ts
anna-carroll: https://github.com/celo-org/optics-monorepo/pull/448/commits/2f7343bf7058c5d05ee8990a183edfc5e3893509

solidity/optics-core/package.json
anna-carroll: https://github.com/celo-org/optics-monorepo/pull/448/commits/923f12a22b0f02e964d73612b92df200850f5744

prestwich:
can you also reset to main and remove the committed files that aren't relevant?


```
git checkout main
git pull
git checkout anna/verify
git reset main --soft
git status

# add only relevant files
git add {specific thing}

git commit --no-verify
git push
```

solidity/optics-core/env.js
```diff
@@ -0,0 +1,17 @@
+const envy = require('envy');
```
prestwich, line 1: do we still need this?

anna-carroll: https://github.com/celo-org/optics-monorepo/pull/448/commits/75ca26f76a4693b3dea1c8d76928b8f430d85a82

anna-carroll:
> can you also reset to main and remove the committed files that aren't relevant?

yep removed 🙂 there is still the deploy output and the new typescript, though 👍

typescript/package.json
```diff
@@ -16,7 +16,8 @@
   "description": "Optics typescript tools",
   "main": "src/index.ts",
   "scripts": {
-    "test": "echo \"Error: no test specified\" && exit 1"
+    "test": "echo \"Error: no test specified\" && exit 1",
+    "deploy": "npx ts-node ./scripts/deploy.ts && cd ../solidity/optics-core && npm run verify"
```
prestwich, line 21: relies on a non-existant file

prestwich (changes requested):
can't have an npm script rely on an uncommitted file. everything else is looking good. I ran prettier on all your stuff

prestwich (approved):

Issue #455, opened 2021-07-02
Relayer panic


prestwich:
This is probably related to the `!time` error, but because it occurred on kovan, the revert message was suppressed.

The real question is what caused the panic 🤔 

```
[...]

PR #457, opened 2021-07-02
feature: time out txns that have been pending confirmation for 10m

prestwich:
Adds a timeout to all ethereum txn dispatches. Stops polling the pending txn after 600 seconds to prevent permanent deadlocks in agents.

Related to #455


prestwich:
why 10 minutes? no reason. probably ought to be shorter

prestwich:
related:
https://github.com/gakonst/ethers-rs/pull/327

PR #468, opened 2021-07-06
test: SimpleMessage ts

ErinHales:
- Add return types
 - Fix GovernanceMessage enum error
 - Add SimpleMessage tests in ts


ErinHales:
Going to merge so I can use in Governance Router tests.  Feel free to still comment if you see anything amiss

PR #476, opened 2021-07-08
Prestwich/no ordering

### LINKED ISSUE #344: Think about spam
**problem**
- it's cheaper to send a message than it is to process it 
- anyone can send a bunch of messages to Optics to clog the system up
- we can't really solve this with governance because the GovernanceRouter uses Optics to send messages 

**ideas**
- cap number of messages in the queue?
- blacklist certain senders? (not sybil-resistant)

### LINKED ISSUE #469: Remove sequence ordering guarantees from replicas
related to #344 

Replicas currently enforce ordering by sequence. However, this allows a sender to add many leaves to the Home tree to cause high processing costs on the Replica and squeeze out other messages.

To prevent this we want to remove the sequence requirements from the replica. This has large implications for the processor, which relies on the `nextToProcess` 

Ideas:
1. simply disable the sequencing requirement. Keep `nextToProcess` but allow it to skip numbers.  Add filtering capabilities to the processor e.g. sender allow/block lists
2. Remove `nextToProcess` as a variable. Refactor the process (how?) to handle this

I'm leaning towards 1. This minimizes the refactoring costs for the processor (at the cost of ~10k gas per message received)

prestwich:
Remove sequence ordering requirement from replica processing

closes #344 
closes #469


solidity/optics-core/contracts/Replica.sol
```diff
@@ -39,13 +39,24 @@ contract Replica is Initializable, Common, QueueManager {
     /// @notice Mapping of enqueued roots to allowable confirmation times
     mapping(bytes32 => uint256) public confirmAt;
 
+    /// @dev re-entrancy guard
+    uint8 private entered = 1;
```
anna-carroll, line 43: the storage variable `entered` is set to 1 on the implementation, but not set on the proxy. so `process` will incorrectly reject with `!reentrant` because `entered = 0` on the proxy. (tests are failing with this)

can we inherit OpenZep `ReentrancyGuard.sol` from [their contracts-upgradable library](https://github.com/OpenZeppelin/openzeppelin-contracts-upgradeable/blob/master/contracts/security/ReentrancyGuardUpgradeable.sol) in order to be compatible with the proxy setup?

rust/processor/src/processor.rs
```diff
@@ -95,6 +106,20 @@ impl ReplicaProcessor {
 
                 let message = message.unwrap();
 
+                // check allow/deny lists
+
+                // if we have an allow list, filter senders not on it
+                if let Some(false) = self.allowed.as_ref().map(|set| set.contains(&message.message.sender)) {
```
anna-carroll, line 112: can you help me think through the setting where we'd want to use an `allowed` list rather than a `denied` list? it seems very restrictive?

solidity/optics-core/contracts/Replica.sol
anna-carroll: alternatively: we could set `entered = 1` in `initialize` if we were strongly against adding an additional dependency, but i'd vote for leveraging OpenZep's existing work

anna-carroll (changes requested):

solidity/optics-core/contracts/Replica.sol
prestwich: I'd like to skip the extra dep, as we use it for only 1 function

rust/processor/src/processor.rs
prestwich: well, because sequencing can now be out of order, it opens up the doors for application-specific processors. e.g. this processor ONLY processes bridge messages. an `allowed` list is the best way to implement that

anna-carroll: ahhhh i see. very cool 😃

anna-carroll (dismissed):
lgtm 🎂

solidity/optics-core/contracts/Replica.sol
```diff
@@ -39,13 +39,24 @@ contract Replica is Initializable, Common, QueueManager {
...
+
```
anna-carroll, line 44: no longer necessary to set `entered = 1` in the implementation

PR #485, opened 2021-07-09
Peg prettier to specific version

anna-carroll:


anna-carroll:
**Note**: I believe there's a diff in `optics-xapps` Solidity files because I was the last person to run `prettier` there with my incompatible version 🙂

anna-carroll:
closing until we verify the `prettier` version is the problem & determine which version to peg to

PR #491, opened 2021-07-15
test: add replica test case of processing message sent to a non-existent contract address

### LINKED ISSUE #440: test: Replica call recipient that isn't a contract

YoduYodu:
fix #440

### DIFF
--- a/typescript/optics-tests/test/replica.test.ts
+++ b/typescript/optics-tests/test/replica.test.ts
@@ -397,6 +397,25 @@ describe('Replica', async () => {
     );
   });
 
+  it('Fails to process message sent to a non-existent contract address', async () => {
+    const sequence = await replica.nextToProcess();
+    const body = ethers.utils.formatBytes32String('message');
+
+    const opticsMessage = optics.formatMessage(
+      remoteDomain,
+      opticsMessageSender.address,
+      sequence,
+      localDomain,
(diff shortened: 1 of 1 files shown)


typescript/optics-tests/test/replica.test.ts
```diff
@@ -397,6 +397,25 @@ describe('Replica', async () => {
...
+      opticsMessageSender.address,
+      sequence,
+      localDomain,
+      '0x1234567890123456789012345678901234567890', // non-existent contract address
+      body,
+    );
+
+    // Set message status to MessageStatus.Pending
+    await replica.setMessagePending(opticsMessage);
+    await replica.process(opticsMessage);
```
ErinHales, line 415: Remove this line. It's causing the message to be processed twice which is why it's reverting in the line below.

typescript/optics-tests/test/replica.test.ts
```diff
@@ -397,6 +397,25 @@ describe('Replica', async () => {
     );
   });
 
+  it('Fails to process message sent to a non-existent contract address', async () => {
```
ErinHales, line 400: Change to 'Processes message sent to a non-existent contract address'

ErinHales (changes requested):

ErinHales (approved):

PR #495, opened 2021-07-16
fix: sooth typechain gen

YoduYodu:
fix: #471 

### Description
The main idea is to remove generated typechain files from repo all together, by modifying `post-checkout`, `post-merge` git hooks.

* 🔕 When `git pull`, and contracts changed. (`git pull` = `git fetch` + `git merge`)
* 🔔 When `git switch` or `git checkout`, and contracts changed. (`git switch` triggers `post-checkout`)

----------------
My conclusion of optimal solution is that we manually hardhat typechain locally. (Remove typescript/typechain in github repo as we've discussed)

What I've done is add typechain-gen.sh to post-checkout & post-merge hooks. But they don't work as expected.

The reasoning can be explained by this simple example:
* We have two branches, `main`  & `myBranch`, with `myBranch` modified optics-core contracts.
* If you switch from main to `myBranch`, post-checkout won't be triggered. Because after you checkout to `myBranch`,  `git diff-index` won't return anything, even thought optics-core contracts has changed from `main` to `myBranch`.
Same situation applies to `post-merge` as well if you `git pull`.(Yes, `git pull` triggers `git merge`)

So my idea is to simply have a `typechain-gen.sh` that unconditionally runs `hardhat typechain` regardless of git `diff-index`. And let the developer handle it.


.github/workflows/typechain-generation.yml
```diff
@@ -0,0 +1,34 @@
...
+      # TODO: modulize typechain gen script?
+      - name: Install
+        run: |
+          npm i
+          npm run compile
+          cd ../.. 
+          git add typescript/typechain
+          git add solidity/optics-core/contracts
+          git add solidity/optics-core/libs
+          git add solidity/optics-core/interfaces
```
ErinHales, line 33: Good start.  Let's clean this up a bit. Let's make a script that checks if typechain _needs_ to be generated and then skip the linting.  Should look something like this:
```
# Conditionally compile, update optics-core abis and generate typechain
if ! git diff-index --quiet HEAD -- ./solidity/optics-core; then
    echo "+generating core typechain"
    cd ./solidity/optics-core
    hardhat typechain
    # add typechain
    cd ../..
    git add typescript/typechain
else
    echo "+Skipping core typechain generation"
fi

# Conditionally compile, update optics-xapps abis and generate typechain
if ! git diff-index --quiet HEAD -- ./solidity/optics-xapps; then
    echo "+generating xapps typechain"
    cd ./solidity/optics-xapps
    hardhat typechain
    # add typechain
    cd ../..
    git add typescript/typechain
else
    echo "+Skipping xapps typechain generation"
fi
```

.github/workflows/typechain-generation.yml
```diff
@@ -0,0 +1,34 @@
...
+          git commit -m "Typechain Generation"
```
ErinHales, line 34: I see a couple problems on doing this on pull request approval.
1. It will cause merge conflicts with the base branch
2. Committing after approval dismisses the review as stale.  This could be really annoying because we'd almost always have to approve twice 😅

ErinHales: Also, if we decide to gitignore the typechain, there would be no need to commit the changes because they would be ignored :D

YoduYodu: > Also, if we decide to gitignore the typechain, there would be no need to commit the changes because they would be ignored :D

Right 😅 That's not a good approach

.gitignore
```diff
@@ -4,3 +4,5 @@ test_deploy.env
 rust/tmp_db
 rust/tmp.env
 tmp.env
+
+typescript/typechain
```
ErinHales, line 8: Not sure just ignoring the typechain is the right call. I see a couple ways we can do this:
**gitignore typechain**
We would have to:
1. Generate typechain for local code very regularly (on pull, on branch change, on push, etc).
2. Generate typechain during testing/workflows (add typechain gen to `test` script in package.json).
Pros:
1. We wouldn't have to commit anything from a script or inside of a workflow
Cons:
1. We may miss a case where we should have generated typechain

**Ignore typechain diff in PRs**
We would need to:
1. Programmatically resolve conflicts between the base branch and the PR branch and commit changes
Pros:
1. Generating typechain would happen (I think) less often and be less of a nuisance to developers.
Cons:
1. We would have to commit within a workflow
2. I don't know how to ignore diffs for a folder inside a PR

ErinHales: Actually I don't love either of those.  They both seem prone to error. Can you see if it’s possible to rebase just a single folder?  I think the best course for now may be to just improve our resolve-typechain-conflicts script.  Maybe something like this?
```
// I'm sure this isn't syntactically correct lol
git rebase ./typescript/typechain <target-branch>
git rm -rf typescript/typechain
cd solidity/optics-core 
hardhat typechain
cd ../optics-xapps
hardhat typechain
cd ../..
git add typescript/typechain
git rebase --continue
git push -f origin <current-branch>
```

ErinHales:
Closing as stale, will revisit issue in the future

Issue #502, opened 2021-07-18
fix: make mapping of representation token contracts queryable


anna-carroll:

prestwich:
closed by #503

Issue #463, opened 2021-07-05
emit event when new bridge token is deployed


anna-carroll:
so apps can use on-chain events to determine which tokens have been sent to a given chain

prestwich:
closed by #503

PR #503, opened 2021-07-18
Emit event when token deployed & make token mapping public

anna-carroll:
#463 
#502


solidity/optics-xapps/contracts/bridge/TokenRegistry.sol
```diff
@@ -41,16 +41,22 @@ abstract contract TokenRegistry is XAppConnectionClient {
...
+        bytes32 id,
+        address representation
+    );
+
     // Contract bytecode that will be cloned to deploy
     // new representation token contracts
     address internal tokenTemplate;
 
     // local representation token address => token ID
-    mapping(address => TokenId) internal reprToCanonical;
```
anna-carroll, line 49: consider for the future:
should these be named something more intuitive like `lookupRepresentation` ?

prestwich (approved):

PR #507, opened 2021-07-20
readme: normalize case for optics-xapps dir, add self as rust codeowner

emberian:
tiny fix i noticed in passing


barbaraliau (approved):
LGTM

PR #509, opened 2021-07-28
Bridge Router improvements: Fast Liquidity, Upgradable Tokens

### LINKED ISSUE #322: feature: fast liquidity on BridgeRouter

prestwich:
Adds a mechanism for LPs to provide fast liquidity to incoming messages

closes #322 

Sketch:
- While the message is in flight, the LP submits it to the Router
- The router transfers tokens from the LP to the recipient, less 5 bps
- When the message is eventually processed, the router sends all tokens to the LP instead of the recipient


solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
```diff
@@ -22,12 +22,65 @@ contract BridgeRouter is Router, TokenRegistry {
     using BridgeMessage for bytes29;
     using SafeERC20 for IERC20;
 
+    /// @notice 5 bps hardcoded fee. Can be changed by contract upgrade
```
asaj, line 25: nit: Update PR description to match

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
```diff
@@ -22,12 +22,65 @@ contract BridgeRouter is Router, TokenRegistry {
...
+        // parse tokenId and action from message
+        bytes29 _msg = _message.ref(0).mustBeMessage();
+        bytes29 _tokenId = _msg.tokenId();
+        bytes29 _action = _msg.action();
+
+        bytes32 _id = _preFillId(_tokenId, _action);
+
+        require(liquidityProvider[_id] == address(0), "!unfilled");
+
+        if (_action.isTransfer()) {
```
asaj, line 73: Why not `require(_action.isTransfer())`?

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
```diff
@@ -22,12 +22,65 @@ contract BridgeRouter is Router, TokenRegistry {
...
     // ======== Constructor =========
 
     constructor(address _xAppConnectionManager)
         TokenRegistry(_xAppConnectionManager)
     {} // solhint-disable-line no-empty-blocks
 
+    // ======== Fast Liquidity System =========
+
+    /// @dev used to identify a token/transfer pair in the prefill LP mapping.
+    /// This approach has a weakness: a user can send 2 identical messages, but
```
asaj, line 46: Thoughts on adding a UID to the message to get around this quirk?

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
asaj: Curious why hardcode this as opposed to building it into the message and allowing users to specify the fee?

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
```diff
@@ -155,17 +210,31 @@ contract BridgeRouter is Router, TokenRegistry {
...
         // get the token contract for the given tokenId on this chain;
         // (if the token is of remote origin and there is
-        // no existing representation token contract, the TokenRegistry will deploy a new one)
+        // no existing representation token contract, the TokenRegistry will
+        // deploy a new one)
         IERC20 _token = _ensureToken(_tokenId);
+        address _recipient = _action.evmRecipient();
+
+        // If an LP has prefilled this request, the LP gets paid instead of the
+        // recipieint
```
asaj, line 219: nit: typo

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
prestwich: why add complexity and length to every message when not doing that works 90% as well? We can always upgrade to a more complex system later if we figure out it's necessary

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
prestwich: no particular reason

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
prestwich: adds 7000ish gas overhead to all transfer messages, as you have to store and increment a nonce for each message. Seems like the corner case `(same user sends same amount of same token)` isn't worth solving if it imposes costs on all users

prestwich: have a "just don't do that" guard in the UI 🤷‍♀️

asaj: Why do you need to store an increment a nonce, can't you just let the user specify a salt?

prestwich: because users can specify the same salt as some other user and frontrun that other user's message admission, ensuring the other user can't access fast liquidity

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
```diff
@@ -22,12 +22,65 @@ contract BridgeRouter is Router, TokenRegistry {
...
 
+    /// @notice 5 bps hardcoded fee. Can be changed by contract upgrade
+    uint256 public constant PRE_FILL_FEE_NUMERATOR = 9995;
+    uint256 public constant PRE_FILL_FEE_DENOMINATOR = 10000;
+
+    /// @notice A mapping that stores the LP that pre-filled a token transfer
+    /// message
+    mapping(bytes32 => address) public liquidityProvider;
+
+    function _applyPreFillFee(uint256 _amnt) internal pure returns (uint256) {
```
anna-carroll, line 33: we've applied a standard ordering for functions across all of our contracts [according to the Solidity style guide](https://docs.soliditylang.org/en/v0.5.3/style-guide.html#order-of-functions) - so internal functions would go at the bottom of the contract

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
anna-carroll: a `Transfer` action doesn't record the message **_sender_** (`from`), only the message **_recipient_** (`to`).  So, by not including a UUID in the message, we are not blocking a single sender from making multiple transfers of the same token & amount with fast liquidity; we are blocking a single recipient. 

IMO this seems like a larger problem. It seems likely to me that recipient collision would be even more likely than sender collision, and is totally out of the control of the message sender

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
anna-carroll: imo would be better to revert with informative message than to no-op if someone tries to call this with a non-transfer message

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
```diff
@@ -22,12 +22,65 @@ contract BridgeRouter is Router, TokenRegistry {
...
+        bytes29[] memory _views = new bytes29[](2);
+        _views[0] = _tokenId;
+        _views[1] = _action;
+        return TypedMemView.joinKeccak(_views);
+    }
+
+    /// @notice Allows a liquidity provider to pre-fill an incoming transfer
+    /// message. The liquidity provider provides the message in advance, and
+    /// the user receives the amount, less the LP's fee immediately.
+    function preFill(bytes calldata _message) external {
```
anna-carroll, line 63: i think liquidity providers should only be able to pre-fill messages where the sender has signaled that they want fast liquidity. 

imo it will not be unusual for folks to move large amounts of liquidity across chains at a given time. imagine you're trying to move $10M worth of tokens across chains for a use-case that is not time sensitive, and in order to use this bridge you are forced to pay $500k for fast liquidity that you don't need (or at least, forced to run the risk of paying)

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
anna-carroll: to fix this, we can have the LP pass in `sender` as well as `message` to the `preFill` function, and add `sender` to the `prefillId` - this yields an even better result, where duplicate fast liquidity is only blocked for the same `token && sender && recipient && amount`

anna-carroll (commented):
👏 amazing start - should I start working on these changes, or you @prestwich ?

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
prestwich: the problem there is that `sender` will alway be the remote bridge router in this context, not the user who initiated the action. so it can't be used to disambiguate anything

will update the docs to specify that it's independent of the user tho 🤔

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
prestwich: fast liquidity v2

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
prestwich: can you make this change?

prestwich:
needs tests

prestwich:
this now includes #524 

1. note that this REMOVES the BridgeRouter initialzier and makes the TokenRegistry initializer public. Given that the router init did nothing besides call the registry init, this seems cleaner
2. to avoid contract size limits, this changes the initializer to take a pre-deployed beacon, rather than having the initializer deploy the beacon and implementation.
3. This now includes a fast liquidty test
4. this now includes a bugfix for the bridgemessage library

solidity/optics-xapps/contracts/bridge/test/TestFastLiquidity.sol
```diff
@@ -0,0 +1,97 @@
...
+
+        require(t.balanceOf(address(this)) == 10000, "!fee");
+
+        bytes memory _msg;
+        bytes29 _tokenId;
+        bytes29 _action;
+        (_msg, _tokenId, _action) = getMessage();
+        BridgeRouter(address(this)).preFill(_msg);
+
+        require(t.balanceOf(address(this)) == FEE, "!fee");
```
anna-carroll, line 89: okay this test is somewhat confusing to me because the liquidity provider and the `BridgeRouter` are the same contract address

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
anna-carroll: done

anna-carroll (approved):
LGTM!

solidity/optics-xapps/contracts/bridge/test/TestFastLiquidity.sol
prestwich: Yeah, it's a pretty lazy test. It can be broken out into typescript, at the expense of many more lines of code

prestwich:
rust failure is caused by updated toolchain version adding new lints. it's not a functional issue. will be resolved by #516

anna-carroll (approved):
WOOOHOOOOO LGTM 🎉

PR #514, opened 2021-07-30
add: deploy and enroll new chain

ErinHales:
- [ ] add test

### DIFF
--- a/typescript/optics-deploy/src/core/index.ts
+++ b/typescript/optics-deploy/src/core/index.ts
@@ -575,7 +575,7 @@ export async function deployNChains(chains: CoreDeploy[]) {
 export async function deployAndEnrollNewChain(chains: Deploy[], newChain: Deploy) {
   await deployOptics(newChain);
 
-  chains.forEach((chain, i) => {
+  chains.forEach(async (chain, i) => {
     await enrollRemote(chain, newChain);
     await enrollRemote(newChain, chain);
 
--- a/typescript/optics-deploy/src/core/index.ts
+++ b/typescript/optics-deploy/src/core/index.ts
@@ -564,6 +564,34 @@ export async function deployNChains(chains: CoreDeploy[]) {
 }
 
 /**
+ * Deploys Optics on a new chain and enrolls it
+ * in a set of existing deploys.
+ *
+ * @dev The first chain in the sequence will be the governing chain
+ *
+ * @param chains - An array of chain deploys
+ * @param newChain - A fresh deploy for the new chain
+ */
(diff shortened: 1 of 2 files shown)


typescript/optics-deploy/src/deployOptics.ts
```diff
@@ -566,7 +566,7 @@ export async function deployNChains(chains: Deploy[]) {
 export async function deployAndEnrollNewChain(chains: Deploy[], newChain: Deploy) {
   await deployOptics(newChain);
 
-  chains.forEach((chain, i) => {
+  chains.forEach(async (chain, i) => {
```
prestwich, line 569: should this be

` await Promise.all(chains.map(async (chain, i) => { ... }));`

typescript/optics-deploy/src/core/index.ts
```diff
@@ -563,6 +563,34 @@ export async function deployNChains(chains: CoreDeploy[]) {
...
+ *
+ * @param chains - An array of chain deploys
+ * @param newChain - A fresh deploy for the new chain
+ */
+export async function deployAndEnrollNewChain(chains: Deploy[], newChain: Deploy) {
+  await deployOptics(newChain);
+
+  await Promise.all(chains.map(async (chain, i) => {
+    await enrollRemote(chain, newChain);
+    await enrollRemote(newChain, chain);
```
prestwich, line 580: hmm, will this work if the previous deploy ownership has been transferred to governance?

anna-carroll: no, it wont :(

typescript/optics-deploy/src/core/index.ts
```diff
@@ -563,6 +563,34 @@ export async function deployNChains(chains: CoreDeploy[]) {
...
+
+    if (i === 0) {
```
anna-carroll, line 582:
```
    if (i === 0) {
      await transferGovernorship(chain, newChain);
    }

    await relinquish(newChain);
```
this should be done outside of the `chains.map`. we are attempting to `relinquish` ownership of the `newChain` N times, where N = `chains.length` - this won't work

outside of the `map` statement,
```
      Deploy governorChain = chains[0];
      await transferGovernorship(governorChain, newChain);

      await relinquish(newChain);
```

typescript/optics-deploy/src/core/index.ts
anna-carroll: in order to enroll the `newChain` in each existing `chain`, each existing chain will need the `governor` to send that transaction.

anna-carroll (changes requested):
👯‍♀️

prestwich:
closing this as requiring further consideration

PR #521, opened 2021-08-03
Fresh deploy config

anna-carroll:


prestwich:
have there been core contract changes that require us to point agents at a new deploy?

PR #524, opened 2021-08-03
Upgradable BridgeToken contracts

### LINKED ISSUE #482: feature: upgradable BridgeToken implementation
- add ability to upgrade implementation of BridgeToken 
- use UpgradeBeacon pattern so all tokens are upgraded with one transaction

prestwich:
based on #509 
closes #482 

- note that this REMOVES the BridgeRouter initialzier and makes the TokenRegistry initializer public. Given that the router init did nothing besides call the registry init, this seems cleaner
- to avoid contract size limits, this changes the initializer to take a pre-deployed beacon, rather than having the initializer deploy the beacon and implementation.
- This now includes a fast liquidty test
- this now includes a bugfix for the bridgemessage library


prestwich:
note that this REMOVES the `BridgeRouter` initialzier and makes the `TokenRegistry` initializer public. Given that the router init did nothing besides call the registry init, this seems cleaner

typescript/optics-deploy/src/bridge/index.ts
```diff
@@ -19,26 +31,26 @@ export type BridgeDeployOutput = {
  */
 export async function deployBridges(deploys: BridgeDeploy[]) {
   // deploy BridgeRouters
```
anna-carroll, line 33: need to call `deployTokenProxys` here

solidity/optics-xapps/contracts/bridge/TokenRegistry.sol
```diff
@@ -48,9 +51,9 @@ abstract contract TokenRegistry is Initializable, XAppConnectionClient {
         address indexed representation
     );
 
-    // Contract bytecode that will be cloned to deploy
-    // new representation token contracts
-    address internal tokenTemplate;
+    /// @dev The UpgradeBeacon that new tokens proxies will read implementation
+    /// from
+    address public beacon;
```
anna-carroll, line 56: rename `tokenBeacon`

PR #516, opened 2021-08-02
refactor: add aws signer

prestwich:
1. remove unused yubi/ledger code
2. remove unused bin
3. add aws signer
4. add config for aws signer
5. refactor signer instantiation from config to be async
6. deal with the consequences

PR #529, opened 2021-08-04
Demonstrate failing Rust lint

anna-carroll:


prestwich:
This was caused by the CI updating its toolchain version, and is addressed in #516

@emberian if you get time, please spend it on looking at #516 :)

PR #539, opened 2021-08-06
fix: ethers version and generate typechain

ErinHales:
- Updates to latest version of ethers
 - Generates typechain
 - Fixes bug in bridge deploy


ErinHales:
Closing in favor of #541

PR #544, opened 2021-08-09
Refactor: BridgeMessage use message identifier

anna-carroll:


anna-carroll:
This PR got borked somehow during rebases etc, opened a new one #552

PR #564, opened 2021-08-12
Tons of Bridge Tests

prestwich:
tracking PR for ongoing test work


typescript/optics-tests/test/bridge/bridge.test.ts
```diff
@@ -17,7 +21,7 @@ const BRIDGE_MESSAGE_TYPES = {
 
 const typeToBytes = (type: number) => `0x0${type}`;
 
-describe('Bridge', async () => {
+describe.only('BridgeRouter', async () => {
```
anna-carroll, line 24: remove `.only`

typescript/optics-tests/test/bridge/bridge.test.ts
```diff
@@ -57,176 +60,203 @@ describe('Bridge', async () => {
...
-        deployerId,
-      );
+        const representation = await deploy.getTestRepresentation();
+        expect(representation).to.not.be.undefined;
+        repr = representation!;
 
-      await expect(stealTx).to.be.revertedWith(
-        'ERC20: burn amount exceeds balance',
-      );
+        await expect(handleTx).to.emit(deploy.bridgeRouter!, 'TokenDeployed');
```
anna-carroll, line 95: okay i am slightly confused about how this passes - doesn't this execute `handleTx` twice? in which case, the tx on line 84 would emit the event and the tx on 95 would not?

this is passing tho so i guess it somehow runs the same tx

anna-carroll (commented):
lookin good

typescript/optics-tests/test/bridge/bridge.test.ts
prestwich: after it's ready for merge

prestwich:
two things to document better
1. MockCore needs to consume a lot of gas to simuilate being home
2. The test `PROCESS_GAS`  in `bridge.test.ts` needs to be set LOWER than the Replica `PROCESS_GAS` to ensure that we have a safety margin

anna-carroll (approved):
Amazing!~!

PR #569, opened 2021-08-13
feature: deploy core can distinguish between dev/prod/staging

prestwich:
- Add a dev/staging/prod configuration variable to core configs
- Add an environment variable we can set to choose between them


anna-carroll (approved):
LGTM

PR #592, opened 2021-08-17
drive-by cleanups

emberian:
main changes:
- added rust/agents, moved agents there
- moved rust/optics-ethereum to rust/chains
- ran shellcheck on shell scripts, fixed some nits


emberian:
TODO: sign a commit at the tail. @prestwich, should i re-sign all the interim commits as well?

ErinHales:
> TODO: sign a commit at the tail. @prestwich, should i re-sign all the interim commits as well?

Pretty sure they all need to be signed. This is the easiest way to do it, then you should set up auto signing.

prestwich:
apologies for pushing to main a few times and causing conflicts. have been trying to fix an issue with deployed agents

I can rebase and sign everything if you'd like. otherwise, turning on signing by default and rebasing to latest main will sign all commits

prestwich:
for this to be merged:
1. rebase onto main and resolve conflicts
2. sign commits
3. check with @yourbuddyconner to ensure that it doesn't break any docker/helm work

solidity/optics-core/update_abis.sh
```diff
@@ -1,3 +1,7 @@
-cat artifacts/contracts/Replica.sol/Replica.json| jq .abi > ../../rust/optics-ethereum/abis/Replica.abi.json && \
-cat artifacts/contracts/Home.sol/Home.json| jq .abi > ../../rust/optics-ethereum/abis/Home.abi.json && \
-cat artifacts/contracts/XAppConnectionManager.sol/XAppConnectionManager.json | jq .abi > ../../rust/optics-ethereum/abis/XAppConnectionManager.abi.json
\ No newline at end of file
+#!/bin/sh 
+
+copy() {
+    jq .abi < artifacts/contracts/"$1".sol/$.json > ../rust/chains/optics-ethereum/abis/"$1".abi.json
+}
```
prestwich, line 5:
```
copy() {
    jq .abi < artifacts/contracts/"$1".sol/"$1".json > ../../rust/chains/optics-ethereum/abis/"$1".abi.json
}
```

rust/optics-core/src/lib.rs
```diff
@@ -145,15 +145,258 @@ impl Signer for Signers {
...
 
-#[async_trait]
-trait SignerExt: Signer {
-    async fn sign_message_without_eip_155<S: Send + Sync + AsRef<[u8]>>(
-        &self,
-        message: S,
-    ) -> Result<Signature, <Self as Signer>::Error> {
-        let mut signature = self.sign_message(message).await?;
-        signature.v = 28 - (signature.v % 2);
-        Ok(signature)
```
prestwich, line 156: Looks like a messy rebase mostly resotred an old version of this file.

Here are what you need to do to fix:

- Structs in `lib` ought to deleted as they've been moved to `types/`
- `SignerExt` needs to be restored

prestwich:
tests fail because `processor` and `optics-core` use a shared test vectors file in `vectors/`. They open it with a utility function in `optics-core` which has the relative path hardcoded. Now that path is different for each crate, as processor has moved

emberian:
> tests fail because `processor` and `optics-core` use a shared test vectors file in `vectors/`. They open it with a utility function in `optics-core` which has the relative path hardcoded. Now that path is different for each crate, as processor has moved

what do you think about leaning on cargo for these paths? from the docs:

```
CARGO_TARGET_TMPDIR — Only set when building integration test or benchmark code. This is a path to a directory inside the target directory where integration tests or benchmarks are free to put any data needed by the tests/benches. Cargo initially creates this directory but doesn't manage its content in any way, this is the responsibility of the test code. There are separate directories for debug and release profiles.
```

relative paths drive me insane, i have such a hard time keeping track of cwd in my head.

rust/optics-core/src/test_utils.rs
```diff
@@ -16,9 +16,21 @@ pub struct MerkleTestCase {
     pub expected_root: H256,
 }
 
+/// Find a vector file assuming that a git checkout exists
+// TODO: look instead for the workspace `Cargo.toml`? use a cargo env var?
+pub fn find_vector(final_component: &str) -> PathBuf {
```
emberian, line 21: i ended up doing this instead of anything with a cargo env var.

scripts/test-all.sh
```diff
@@ -6,6 +6,8 @@ set -xe
 
 # update ABIs
 (
+    cd ./solidity
```
emberian, line 9: this was removed by accident during rebase

rust/optics-core/src/test_utils.rs
prestwich: cool. this looks fine. would be good to add a `.take(4)` to limit how deep into the directory structure we can go

prestwich:
relative paths also drive me insane. at some point we should do a cleanup and get rid of all of them

prestwich:
give that checks have all passed, I'll merge this once we get approves from 
- @yourbuddyconner (please review the dockerfile changes and ensure this doesn't affect Helm work)
- @ErinHales (please review the .sh script changes)

rust/chains/README.md
```diff
@@ -0,0 +1,4 @@
+To add a new chain:
+
+    - add a new crate here
```
ErinHales, line 3: Assuming "here" is supposed to have a link?

yourbuddyconner (approved):
Dockerfile changes ok

rust/chains/README.md
```diff
@@ -1,4 +1,4 @@
 To add a new chain:
 
-    - add a new crate here
-    - add a new enum variant [there](optics_base::homes::Homes) # TODO: rich text link from md to live rust
+    - add a new crate to this directory, implementing the appropriate traits.
```
emberian, line 3: @ErinHales this is much less ambiguous!

ErinHales:
Getting these errors when running `cargo clippy`
Screen Shot 2021-08-20 at 3 18 34 PM

pre-commit.sh
```diff
@@ -153,9 +153,10 @@ else
 fi
 
 # Git add abis if updated
-if ! git diff-index --quiet HEAD -- ./rust/optics-ethereum/abis/; then
-    echo '+git add ./rust/optics-ethereum/abis/*'
-    git add ./rust/optics-ethereum/abis/*
+abis_dir=./rust/chains/optics-ethereum/abis/
+if ! git diff-index --quiet HEAD -- ./rust/chains/optics-ethereum/abis/; then
```
ErinHales, line 157: Use `abis_dir` in conditional: `if ! git diff-index --quiet HEAD -- abis_dir; then`

emberian:
> Getting these errors when running `cargo clippy`

@ErinHales i also got those, doing a `rustup update` fixed it for me -- i'm currently using `rust 1.54`. i'm going to commit that as a version pin...

ErinHales (approved):

prestwich (approved):

PR #593, opened 2021-08-17
refactor processor to have nicer behavioural traces

### LINKED ISSUE #580: Trace Usage of CommittedMessage
The CommittedMessage and RawCommittedMessage structs are used when a message has been committed at some leaf index. We can use this to understand the portion of the agent message pipeline post-message commit. We can use the leaf index to correlate spans together, include this value in the Span itself. 

Deliverable: 
Produce spans that trace the usage of [CommittedMessage](https://github.com/celo-org/optics-monorepo/blob/main/rust/optics-core/src/traits/home.rs#L23-L28) and [RawCommittedMessage](https://github.com/celo-org/optics-monorepo/blob/main/rust/optics-core/src/traits/home.rs#L14-L19) structs in the agent codebase. 

Occurrences: 
`raw_committed_message_by_sequence`: https://github.com/celo-org/optics-monorepo/blob/main/rust/optics-ethereum/src/home.rs#L184-L203
`raw_message_by_leaf`: https://github.com/celo-org/optics-monorepo/blob/main/rust/optics-ethereum/src/home.rs#L206-L222
`message_by_sequence`: https://github.com/celo-org/optics-monorepo/blob/main/rust/optics-core/src/traits/home.rs#L78-L88
`message_by_leaf`: https://github.com/celo-org/optics-monorepo/blob/main/rust/optics-core/src/traits/home.rs#L99-L108

`process - processor.rs`: https://github.com/celo-org/optics-monorepo/blob/main/rust/processor/src/processor.rs#L163

emberian:
currently working on:

1. reducing the number of events (lots of extraneous ones?)
2. increasing the number of spans
3. making sure the relations between them make sense
4. adding fields + structured metadata connecting disparate things, esp loading up spans with all the metadata events later reference
5. instrumenting optics-base

Closes #580 #631

### DIFF
--- a/rust/agents/kathy/src/kathy.rs
+++ b/rust/agents/kathy/src/kathy.rs
@@ -76,13 +76,13 @@ impl OpticsAgent for Kathy {
                             body,
                         };
                         info!(
+                            target: "outgoing_messages",
                             "Enqueuing message of length {} to {}::{}",
-                            message.body.len(),
-                            message.destination,
-                            message.recipient
+                            length = message.body.len(),
+                            destination = message.destination,
+                            recipient = message.recipient
                         );
                         home.enqueue(&message).await?;
-                        info!("Enqueue success");
                     }
                     _ => {
                         info!("Reached the end of the static message queue. Shutting down.");
--- a/rust/agents/processor/src/main.rs
+++ b/rust/agents/processor/src/main.rs
@@ -23,6 +23,7 @@ async fn _main() -> Result<()> {
     let settings = Settings::new()?;
     settings.base.tracing.start_tracing()?;
 
+    // TODO: top-level root span customizations?
     let agent = Processor::from_settings(settings).await?;
     agent.run_all().await?;
     Ok(())
--- a/rust/agents/processor/src/processor.rs
+++ b/rust/agents/processor/src/processor.rs
@@ -18,7 +18,6 @@ use tokio::{
 };
 use tracing::{error, info, instrument, instrument::Instrumented, warn, Instrument};
 
-use crate::{prover::Prover, prover_sync::ProverSync, settings::ProcessorSettings as Settings};
 use optics_base::{
     agent::{AgentCore, OpticsAgent},
     cancel_task, decl_agent,
@@ -31,10 +30,7 @@ use optics_core::{
     traits::{CommittedMessage, Common, Home, MessageStatus},
 };
 
-use crate::{
-    prover::{Prover, ProverSync},
-    settings::ProcessorSettings as Settings,
-};
+use crate::{prover::Prover, prover_sync::ProverSync, settings::ProcessorSettings as Settings};
 
 /// The replica processor is responsible for polling messages and waiting until they validate
 /// before proving/processing them.
@@ -66,11 +62,12 @@ impl UsingPersistence<usize, Proof> for Replica {
     }
 }
 
-impl UsingPersistence<usize, u32> for Replica {
+// 'static usually means "string constant", don't dynamically create db keys
+impl UsingPersistence<&'static str, u32> for Replica {
     const KEY_PREFIX: &'static [u8] = b"state_";
 
-    fn key_to_bytes(key: usize) -> Vec<u8> {
-        key.to_be_bytes().into()
+    fn key_to_bytes(key: &'static str) -> Vec<u8> {
+        key.into()
     }
 }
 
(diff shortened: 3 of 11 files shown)


rust/processor/src/processor.rs
```diff
@@ -89,76 +119,86 @@ impl ReplicaProcessor {
...
+                    info!(
+                        "Dispatching a message for processing {}:{}",
+                        domain, sequence
+                    );
+                    self.process(message, proof).await?;
+                    Ok(LoopAction::DoSleep)
+                }.instrument(seq_span);
+
+                let cfa = control_flow_action.await?;
+                // TODO: ought this not be updated if SkipSleep?
```
emberian, line 188: @prestwich can you help me repair the control flow here? (i hate upgrading second class control flow to first class...). i'm overal confused about the timed loop.

prestwich: Which part of the control flow do you think is broken?

The purpose of the timed loop is to 
a) walk the agent through the process of getting the next message, and ensuring it has the info necessary to process it
b) rate limit access to the chain RPC so we don't overload nodes or get API access revoked

Generally, the actions described in the comment at the top of the loop are correct and should be represented in the the internal flow. This loop is a bit of a kludge. Can we start by breaking it down into smaller pieces that represent those actions?

rust/processor/src/processor.rs
```diff
@@ -76,131 +47,141 @@ pub(crate) struct ReplicaProcessor {
...
-                        sequence += 1;
-                        reset_loop!(self.interval);
-                    }
+    /// Returns the next sequence number to try and process, for the given domain.
+    ///
+    /// Postcondition: if retval? != current_seq { message was filtered ⊻ message was processed }
+    ///                else { assert retval? == current_seq; prerequisite data not yet found }
+    /// In case of error: send help?
+    #[instrument(err)]
+    async fn try_msg_by_domain_and_seq(&self, domain: u32, current_seq: u32) -> Result<u32> {
```
emberian, line 115: @prestwich what do you think about this restructuring?

emberian:
@yourbuddyconner i'm almost ready to sit down and figure out spanyeet, the processor instrumentation should Make Sense now and be a good example for the pattern to push into the rest of the code

rust/processor/src/processor.rs
prestwich: hmmm so you're making me re-examine how this was designed in the first place and I'm seeing more holes

z`Replica::next_to_process` was intended to be a hint on where in the message set the processor should start looking for unprocessed messages. However, if we retrieve it from the chain, it's possible that we end up skipping local processing of a batch of messages because some other processor has submitted a later message

e.g. 
- messages 0-50 arrive in an `Update`
- Processor A processes ONLY message 46
- `nextToProcess` is set to 47
- Processor B starts processing at 47, _regardless of whether any unprocessed messages exist in 1-45_

Proposed fix:
- Delete `nextToProcess` on Replica.sol
- `Processor` inspects _each_ message
- `Processor` persists the sequence of the last inspected message in its DB as `lastInspected`
- on agent startup, processor pulls `lastInpected` from its DB and starts processing at that number

Thoughts

rust/optics-ethereum/src/replica.rs
```diff
@@ -213,11 +213,6 @@ where
         Ok(report_tx!(tx).into())
     }
 
-    #[tracing::instrument(err)]
```
prestwich, line 216: PR needs to be marked as closing:
#619 #620 #621 

cc @anna-carroll

rust/optics-core/src/lib.rs
```diff
@@ -266,6 +266,16 @@ pub struct Update {
     pub new_root: H256,
 }
 
+impl core::fmt::Display for Update {
```
prestwich, line 269: will need a rebase as this struct has moved already in `main`. but wait until after #592 is merged

rust/processor/src/processor.rs
```diff
@@ -46,121 +47,151 @@ pub(crate) struct ReplicaProcessor {
...
-                    sequence,
-                );
-
-                let proof = proof_opt.unwrap();
-                if proof.leaf != message.to_leaf() {
-                    let err = format!("Leaf in prover does not match retrieved message. Index: {}. Calculated: {}. Prover: {}.", message.leaf_index, message.to_leaf(), proof.leaf);
-                    error!("{}", err);
-                    bail!(err);
-                }
+    /// Returns the next sequence number to try and process, for the given domain.
```
prestwich, line 119: No longer true. Now returns a boolean indicating whether the message should be marked as inspected (i.e. whether the exterior loop should make progress or run again

rust/processor/src/processor.rs
```diff
@@ -46,121 +47,151 @@ pub(crate) struct ReplicaProcessor {
...
+                // 5. Submit the proof to the replica
+                let mut next_to_inspect: u32 = match Self::db_get(&self.db, "lastInspected") {
+                    Some(n) => n + 1,
+                    None => 0,
+                };
+                loop {
+                    let seq_span = tracing::trace_span!(
+                        "ReplicaProcessor",
+                        name = self.replica.name(),
+                        sequence = next_to_inspect
```
prestwich, line 92: maybe add domain informaiton here? `self.replica.domain()` and `self.home.domain()` ?

rust/processor/src/processor.rs
prestwich: so `Ok(true)` means either "skip this message" or "this message was processed"

prestwich: while `Ok(false)` means "try again later

rust/processor/src/processor.rs
```diff
@@ -46,121 +47,151 @@ pub(crate) struct ReplicaProcessor {
...
+            error!("{}", msg);
+            bail!(msg);
+        }
+
+        while !self.replica.acceptable_root(proof.root()).await? {
+            info!(
+                "Proof under root {root} not yet valid here",
+                root = proof.root(),
+            );
+            sleep(Duration::from_secs(self.interval)).await;
```
prestwich, line 178: should add a comment on this behavior. Once we've established that a message exists and we have a proof, we wait here until the update containing it is confirmed on the `Replica`

prestwich (changes requested):

prestwich:
needs rebase onto main after #629 

- solidity changes should be dropped from this PR

rust/processor/src/processor.rs
emberian: i think of `info!()` as an executable comment :) tweaking its prose.

prestwich (approved):

PR #595, opened 2021-08-18
Helm Chart Update

ajcrowe:
This updates the Helm chart with a few extra params and simplifies some of the common configuration

* Updated to use dedicated resources and pod lables/annotations
* Added ConfigMap for common envs
* Updated storage configuration to be more flexible
* Added missing envs for db path, backtrace and loggings

### DIFF
--- a/rust/helm/optics-agent/templates/_helpers.tpl
+++ b/rust/helm/optics-agent/templates/_helpers.tpl
@@ -35,7 +35,6 @@ Common labels
 */}}
 {{- define "optics-agent.labels" -}}
 helm.sh/chart: {{ include "optics-agent.chart" . }}
-optics/deployment: {{ .Values.optics.runEnv | quote }}
 {{ include "optics-agent.selectorLabels" . }}
 {{- if .Chart.AppVersion }}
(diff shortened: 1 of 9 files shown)


yourbuddyconner:
Tested this change and it works great, thanks for the contribution @ajcrowe!

yourbuddyconner (approved):

rust/helm/optics-agent/templates/_helpers.tpl
```diff
@@ -35,7 +35,6 @@ Common labels
 */}}
 {{- define "optics-agent.labels" -}}
 helm.sh/chart: {{ include "optics-agent.chart" . }}
-optics/deployment: {{ .Values.optics.runEnv | quote }}
```
yourbuddyconner, line 38: @ajcrowe We actually use this label in our dashboards to filter logs from a particular deployment. Do you have any suggestions on how we can preserve this functionality?

prestwich:
failing checks are because of recent changes to CI workflows. Will rerun

prestwich:
ah correction, checks are failing due to #624

PR #616, opened 2021-08-19
refactor: split cache, install independently

ErinHales:
I know it's verbose, but it will save us time on checks

### DIFF
--- a/.github/workflows/solidity.yml
+++ b/.github/workflows/solidity.yml
@@ -14,75 +14,89 @@ jobs:
   install:
     runs-on: ubuntu-latest
 
-    defaults:
-      run:
-        working-directory: ./solidity
-
     steps:
       - uses: actions/checkout@v2
 
-      - uses: actions/cache@v2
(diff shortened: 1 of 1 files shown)


emberian:
i ask out of ignorance and curiosity: are these caches distributable artifacts? do they get baked into containers at some point?

ErinHales:
> i ask out of ignorance and curiosity: are these caches distributable artifacts? do they get baked into containers at some point?

Hey!  So they are not artifacts.  I'm not sure how GitHub handles the cache behind the scene. Here's a [link](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows) with more information if you want to read up on it more :)

prestwich (approved):

ErinHales:
The reason for this change is it takes a long time to install dependencies in all the folders.  And currently the cache is batched together, meaning if there's a single dependency change it will have to rerun and recache the entire thing

PR #642, opened 2021-08-25
use OTLP for exporting spans

emberian:
current status:

1. confirmed that spans are making it to grafana agent
2. seems to work fine over plaintext tcp, who needs authentication or integrity when you're inside a k8s cluster?
3. cannot actually find any of the purported spans when looking at the grafana explorer.
4. the behavior of the tracing is bizarre, cannot explain why, for kathy, `OPT_BASE_TRACING_LEVEL=info` outputs a different set of `info` traces than `OPT_BASE_TRACING_LEVEL=trace`. there should be a partition relation here, but for some reason there's just a subset.
5. the sampling status and tracing level needs to be a tweakable, doing it statically is hilariously verbose and very naive.

Progress on #449


rust/optics-base/src/settings/trace/mod.rs
```diff
@@ -62,8 +57,7 @@ impl From<Level> for LevelFilter {
 /// Configuration for the tracing subscribers used by Optics agents
 #[derive(Debug, Clone, serde::Deserialize)]
 pub struct TracingConfig {
-    jaeger: Option<JaegerConfig>,
-    zipkin: Option<ZipkinConfig>,
+    otlp_endpoint: Option<String>,
```
prestwich, line 60: remove `_` from name as it causes issues with setting config by ENV var

rust/optics-base/src/settings/trace/mod.rs
```diff
@@ -81,17 +75,23 @@ impl TracingConfig {
...
+        match self.otlp_endpoint {
+            None => subscriber.try_init()?,
+            Some(ref uri) => {
+                let tracer = opentelemetry_otlp::new_pipeline()
+                    .tracing()
+                    .with_exporter(
+                        opentelemetry_otlp::new_exporter()
+                            .tonic()
+                            .with_endpoint(uri)
+                            .with_timeout(std::time::Duration::from_secs(3)),
```
prestwich, line 87: I don't understand any of this so I will rely on you and @yourbuddyconner here

Issue #657, opened 2021-08-26
Prometheus Metrics Implementation


prestwich:
Epic container for Agent Prometheus Metrics

Relevant Discussion: https://github.com/celo-org/optics-monorepo/discussions/645

emberian:
>  metrics mod contains once_cell for a CoreMetrics struct

what do you mean by this? the global state lives under this namespace?

emberian:
https://prometheus.io/docs/concepts/data_model/

emberian:
wallet address doesn't make sense to me as a metric.

prestwich:
> > metrics mod contains once_cell for a CoreMetrics struct
> 
> what do you mean by this? the global state lives under this namespace?

metrics that are common across all agents should be defined in a struct in `optics_base::metrics`. If we want to make these into global state, use `once_cell`, not `lazy_static` ( [see here for example](https://docs.rs/prometheus/0.12.0/prometheus/#static-metrics) )

prestwich:
> wallet address doesn't make sense to me as a metric.

not really, but it's kinda useful to display them alongside the metrics

emberian:
# metric naming

Using prometheus terminology, metrics have a standard format of `namespace_subsystem_name_units`. `optics` will be the `namespace`. Should the agents have their own metrics, making `subsystem` an agent name, or should the particular agent instead be a `label`? Labels are a way of slicing up datapoints recorded under a single metric along arbitrary dimensions, specified in the opentsdb format inside braces, such as `optics_updater_wallet_balance_total` vs `optics_wallet_balance_total{agent="updater"}`.

emberian:
# datatype compatibility

prometheus only supports double and uint64, excluding in full generality wallet balance from being a metric datapoint. should we truncate to 64 bits? convert to float?

PR #670, opened 2021-08-29
rust agents: expose prometheus metrics

### LINKED ISSUE #717: metrics: Implement Metrics Scaffolding
here's the implementation sketch

- [ ] Add a `metrics` mod to `optics-base`
- [ ] metrics mod contains once_cell for a `CoreMetrics` struct
- [ ] `CoreMetrics` contains a selection of metrics and exposes a `gather`
    - tx counts
    - wallet balance
    - blockheight
    - wallet address 
- [ ] Define an `AgentMetrics` in each agent with `once_cell` in a `metrics.rs`
    - macroify (see `decl_settings` macro)
- [ ] `AgentsMetrics` exposes agent-specific metrics and exposes a `gather`
    - e.g. updater attestations signed and total
    - e.g. current `next_to_inspect` for processor

_Originally posted by @prestwich in https://github.com/celo-org/optics-monorepo/discussions/645#discussioncomment-1241232_

emberian:
Progress on #657
Closes #717

### DIFF
--- a/rust/agents/updater/src/main.rs
+++ b/rust/agents/updater/src/main.rs
@@ -22,6 +22,9 @@ async fn _main() -> Result<()> {
     settings.base.tracing.start_tracing()?;
 
     let agent = Updater::from_settings(settings).await?;
+
+    let _ = agent.metrics().run_http_server();
+
     agent.run_all().await?;
     Ok(())
 }
--- a/rust/optics-base/src/settings/mod.rs
+++ b/rust/optics-base/src/settings/mod.rs
@@ -143,6 +143,8 @@ impl SignerConf {
 pub struct Settings {
     /// The path to use for the DB file
     pub db: String,
+    /// Port to listen for prometheus scrape requests
+    pub metricsport: Option<u16>,
     /// The home configuration
     pub home: ChainSetup,
     /// The replica configurations
@@ -182,13 +184,22 @@ impl Settings {
         self.home.try_into_home(signer).await
     }
 
-    /// Try to generate an agent core
-    pub async fn try_into_core(&self) -> Result<AgentCore, Report> {
+    /// Try to generate an agent core for a named agent
+    pub async fn try_into_core(&self, name: &str) -> Result<AgentCore, Report> {
         let home = Arc::new(self.try_home().await?);
         let replicas = self.try_replicas().await?;
         let db = Arc::new(db::from_path(&self.db)?);
 
-        Ok(AgentCore { home, replicas, db })
+        Ok(AgentCore {
+            home,
+            replicas,
+            db,
+            metrics: Arc::new(crate::metrics::CoreMetrics::new(
+                name,
+                self.metricsport,
+                Arc::new(prometheus::Registry::new()),
+            )?),
+        })
     }
 
     /// Read settings from the config file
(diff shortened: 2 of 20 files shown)


emberian:
@prestwich this isn't quite what you spec'd out, when it comes to the agent details, but i think it's lower boilerplate. how do you think this looks?

rust/agents/updater/src/main.rs
```diff
@@ -22,6 +23,17 @@ async fn _main() -> Result<()> {
...
 
     let agent = Updater::from_settings(settings).await?;
+
+    let metrics = agent.metrics();
+    let metrics_endpoint =
+        warp::path!("metrics").map(move || metrics.gather().expect("failed to encode metrics"));
+
+    tokio::spawn(async move {
+        warp::serve(metrics_endpoint)
+            .run(([0, 0, 0, 0], 8080))
```
yourbuddyconner, line 33: Please use port 9090 by default here, allowing it to be set using the config pipeline would be nice too.

emberian:
afaict the only major problem with this right now is running multiple agents in the same binary is unreasonable (can't merge the registries/HTTP servers).

yourbuddyconner:
> afaict the only major problem with this right now is running multiple agents in the same binary is unreasonable (can't merge the registries/HTTP servers).

I dont think we anticipate ever running multiple instances of an agent from a single binary, so this is an acceptable constraint.

prestwich:
I am generally liking what I see. Can we run from this branch and confirm that the endpoint works?

rust/optics-base/src/settings/mod.rs
```diff
@@ -196,7 +196,9 @@ impl Settings {
             db,
             metrics: Arc::new(crate::metrics::CoreMetrics::new(
                 name,
-                self.metrics,
+                self.metrics
+                    .as_ref()
+                    .map(|v| v.parse::<u16>().expect("metrics port must be u16")),
```
prestwich, line 201: can we do a reasonable default and `unwrap_or` ?  

cc @yourbuddyconner what would a reasonable default metrics port be?

yourbuddyconner: Port 9090 is default prometheus port

emberian: i would rather have a way to disable the metrics server.

the reasonable default is in the default config, is that a bad place for it?

prestwich: I am hard pressed to think of a situation where we would want to disable metrics. Do you have a use case in mind?

emberian: i probably wouldn't use metrics on my own xapp agent and until this PR i never got the popup from macOS to punch a hole through the firewall for a listening port. not particularly beyond operator paranoia.

yourbuddyconner:
Screen Shot 2021-09-03 at 12 33 26 PM

Verified that this endpoint loads in a dev environment, haven't tested whether metrics actually increment. We can probably merge this scaffolding and then verify individual metrics in subsequent PRs.

yourbuddyconner:
This PR closes #717

rust/optics-base/src/settings/mod.rs
yourbuddyconner: Should definitely be switchable. Default should be off, and we should include a default config for port 9090 in the config generator as we will never run agents without metrics, and neither will anyone we know.

yourbuddyconner:
Also verified that prometheus scrapes this endpoint alright. Think this scaffolding PR is good to go. 

Screen Shot 2021-09-03 at 12 55 38 PM

yourbuddyconner (approved):

PR #671, opened 2021-08-29
bug: transfer root governorship to configured governor on gov chain

prestwich:
1. adds a configuration option and method to transfer the root governorship of the governing chain's governance router on deployment
2. improves the deployment contract output to include other relevant addresses

### DIFF
--- a/typescript/optics-deploy/src/core/index.ts
+++ b/typescript/optics-deploy/src/core/index.ts
@@ -460,6 +460,28 @@ export async function transferGovernorship(gov: CoreDeploy, non: CoreDeploy) {
 }
 
 /**
+ * Installs the intended ultimate governor in that domain's Governance Router.
+ * If the governor address is not configured, it will remain the deployer
+ * address.
+ * @param gov - The governor chain deploy instance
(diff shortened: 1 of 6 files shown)


typescript/optics-deploy/src/core/index.ts
```diff
@@ -501,6 +523,10 @@ export async function deployTwoChains(gov: CoreDeploy, non: CoreDeploy) {
 
   await Promise.all([relinquish(gov), relinquish(non)]);
 
+  if (gov.config.governor) {
+    await appointGovernor(gov);
```
ErinHales, line 527: Where is this `appointGovernor` function coming from? Did you mean to rename `installGovernor`?

ErinHales (changes requested):

typescript/optics-deploy/src/core/index.ts
prestwich: ah yep. I did. thank you. my TS error checker is apparently broken 😬

ErinHales (approved):

PR #697, opened 2021-09-01
Final Contract Cleanup

anna-carroll:


solidity/optics-core/contracts/Replica.sol
```diff
@@ -14,64 +14,83 @@ import {TypedMemView} from "@summa-tx/memview-sol/contracts/TypedMemView.sol";
...
     uint32 public remoteDomain;
-
     /// @notice Number of seconds to wait before enqueued root becomes confirmable
     uint256 public optimisticSeconds;
-
-    /// @notice Mapping of enqueued roots to allowable confirmation times
-    mapping(bytes32 => uint256) public confirmAt;
-
     /// @dev re-entrancy guard
     uint8 private entered;
```
emberian, line 55: this is in the "public storage" section -- should it get shuffled below the other public fields?

PR #698, opened 2021-09-01
Final Contract Cleaning + Renaming

anna-carroll:


rust/optics-core/src/test_output.rs
```diff
@@ -43,7 +43,7 @@ pub mod output_functions {
             "sender": optics_message.sender,
             "destination": optics_message.destination,
             "recipient": optics_message.recipient,
-            "sequence": optics_message.sequence,
+            "nonce": optics_message.nonce,
             "body": optics_message.body,
             "leaf": optics_message.to_leaf(),
```
prestwich, line 48: messageHash

anna-carroll: `"messageHash": optics_message.message_hash` ❓

anna-carroll: https://github.com/celo-org/optics-monorepo/pull/698/commits/6bbccdb8cdd0909487ec29387ab3bed5f86cd7e0

prestwich (approved):

PR #700, opened 2021-09-01
test: deploy scripts

ErinHales:
- asserts that all elements in a deploy are defined
 - asserts that the governor chain has governorship over all other chains
 - asserts that governance router has ownership over other contracts
 - asserts home is set on xAppConnectionManager

TODO:
 - [x] Check deployNChains works with 1 chain
 - [x] Add checkBridgeDeploy


anna-carroll:
Looking good. Can we update our test suite to run against contracts deployed using the N deploy script?

typescript/optics-tests/test/deploy.test.ts
```diff
@@ -0,0 +1,121 @@
...
+
+const domains = [1000, 2000, 3000];
+
+function assertBeaconProxy(beaconProxy: BeaconProxy<types.Contract>) {
+  expect(beaconProxy.beacon).to.not.be.undefined;
+  expect(beaconProxy.proxy).to.not.be.undefined;
+  expect(beaconProxy.implementation).to.not.be.undefined;
+}
+
+export async function checkDeploy(
```
prestwich, line 20: love this, we should have it in the `optics-deploy` package instead of in `optics-test`

prestwich: basically I think this should be part of the official deploy process :)

prestwich: the next steps:

- for an n-chain deploy, check that all chains have enrolled replicas for all other chains
- `checkBridgeDeploy`

ErinHales:
> Looking good. needs conflicts resolved.
> 
> Can we update our test suite to run against contracts deployed using the N deploy script? => in cross-chain tests, replace `deployTwoChains` with `deployNChains`

Yes! Will update that and resolve conflicts

ErinHales:
Update: couple tests failing.  Debugging now

ErinHales:
I think the deploy is fine, just sent an extra message so we need to generate a new proof with the correct index.  I also don't want to do that tho...

ErinHales:
> I think the deploy is fine, just sent an extra message so we need to generate a new proof with the correct index.  I also don't want to do that tho...

Scratch that. I think we may be doing a step twice which results in an extra message being sent

typescript/optics-deploy/src/test/core.ts
```diff
@@ -0,0 +1,77 @@
+import { expect } from 'chai';
```
prestwich, line 1: I think I'd prefer this file to be at `src/core/checks.ts`

prestwich: that way we can later add `src/bridge/checks.ts` and it'll be ocnsistent

ErinHales: Much better, thanks

solidity/optics-xapps/contracts/bridge/test/TestBridgeRouter.sol
```diff
@@ -0,0 +1,10 @@
+// SPDX-License-Identifier: MIT OR Apache-2.0
+pragma solidity >=0.6.11;
+
+import {BridgeRouter} from "../BridgeRouter.sol";
+
+contract TestBridgeRouter is BridgeRouter {
+    function getRemoteRouter(uint32 _domain) external view returns (bytes32) {
```
prestwich, line 7: I opened #750 to replace this test contract

prestwich: Cool you can refactor to simplify now :)

prestwich (dismissed):

prestwich:
ran prettier. will merge when checks pass

ErinHales:
Will resolve conflicts here in a minute

prestwich (dismissed):

prestwich (approved):

PR #702, opened 2021-09-01
feature: add verify scripts for more networks

prestwich:


ErinHales:
As far as I can tell by looking at it, this _should_ work just fine. But would need to test with mainnet deploy

anna-carroll (approved):
LGTM

PR #726, opened 2021-09-06
refactor: update poller loop behavior to reduce latency

prestwich:
- change behavior of poller when queue is full to drop updates
- change queue length to 1
- add a pre-wait check in the update handler's handle function

CC @nambrot thanks :)

### DIFF
--- a/rust/agents/updater/src/updater.rs
+++ b/rust/agents/updater/src/updater.rs
@@ -11,7 +11,7 @@ use prometheus::IntCounter;
 use rocksdb::DB;
 use tokio::{
     sync::{
-        mpsc::{self, Receiver, Sender},
+        mpsc::{self, error::TrySendError, Receiver, Sender},
         Mutex,
     },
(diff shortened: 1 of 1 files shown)


rust/agents/updater/src/updater.rs
```diff
@@ -112,6 +111,12 @@ impl UpdateHandler {
     #[tracing::instrument(err, skip(self), fields(self = %self))]
     async fn handle_update(&self, update: Update) -> Result<()> {
         info!("Have an update, awaiting the tick");
+
+        if !self.acceptable(&update).await? {
+            info!("Declined to submit update. No longer current");
+            return Ok(());
+        }
+
```
anna-carroll, line 119: can we document why we check => sleep => check again? 
check 1: discard the suggested update if it's stale / old
sleep + check 2: wait in case a tx has already been submitted (?) and then check again

anna-carroll (approved):
dope!

PR #727, opened 2021-09-06
feature: make PROCESS_GAS immutable

prestwich:


emberian (approved):
lgtm

solidity/optics-core/contracts/Replica.sol
```diff
@@ -86,7 +86,16 @@ contract Replica is Version0, Common {
     // ============ Constructor ============
 
     // solhint-disable-next-line no-empty-blocks
-    constructor(uint32 _localDomain) Common(_localDomain) {}
+    constructor(
+        uint32 _localDomain,
+        uint256 _processGas,
+        uint256 _reserveGas
+    ) Common(_localDomain) {
+        require(_processGas >= 850_000, "!process gas");
```
anna-carroll, line 94: why underscore here?

anna-carroll (approved):
LGTM

solidity/optics-core/contracts/Replica.sol
anna-carroll: seems like this is acceptable syntax for spacing number literals

prestwich: yes, the syntax was introduce to make numeric literals more legible

PR #735, opened 2021-09-06
bug: metrics no longer panics when attempting to register multiple replicas

prestwich:
privated the metrics registry and added official interface for setting up per-replica metrics


yourbuddyconner (approved):
Validated this change in dev

Issue #723, opened 2021-09-03
metrics: Processor - Next Message to Inspect


yourbuddyconner:
```
# Gauge for next message in the tree to process
optics_next_message_index{agent="processor"}
```

yourbuddyconner:
Implemented as part of #746

PR #746, opened 2021-09-07
feature: additional agent metrics

### LINKED ISSUE #719: metrics: Attestation Count
```
# counter for updater and watcher number of signed attestations
optics_attestation_count{network="<NETWORK>", agent="<AGENT>"}
```

emberian:
Cleans up some existing metrics, using labels for dimensioning.

Closes #719 #723 #718

### DIFF
--- a/rust/agents/processor/src/processor.rs
+++ b/rust/agents/processor/src/processor.rs
@@ -268,7 +268,7 @@ impl Processor {
                 "Index of the next message to inspect",
                 &["replica", "agent"],
             )
-            .expect("Processor metric alreadt registered -- should have be a singleton");
+            .expect("Processor metric already registered -- should have be a singleton");
 
         Self {
             interval,
--- a/rust/agents/processor/src/processor.rs
+++ b/rust/agents/processor/src/processor.rs
@@ -32,7 +32,8 @@ use optics_core::{
 
 use crate::{prover::Prover, prover_sync::ProverSync, settings::ProcessorSettings as Settings};
 
-static LAST_INSPECTED: &str = "lastInspected";
+const LAST_INSPECTED: &str = "lastInspected";
+const AGENT_NAME: &str = "processor";
 
 /// The replica processor is responsible for polling messages and waiting until they validate
 /// before proving/processing them.
@@ -44,7 +45,7 @@ pub(crate) struct Replica {
     db: Arc<DB>,
     allowed: Option<Arc<HashSet<H256>>>,
     denied: Option<Arc<HashSet<H256>>>,
-    next_to_inspect: prometheus::IntGauge,
+    next_message_index: prometheus::IntGaugeVec,
 }
 
 impl std::fmt::Display for Replica {
@@ -91,19 +92,23 @@ impl Replica {
                 //      - If not, wait and poll again
                 // 4. Check if the proof is valid under the replica
                 // 5. Submit the proof to the replica
-                let mut next_to_inspect: u32 = match Self::db_get(&self.db, LAST_INSPECTED)? {
+                let mut next_message_index: u32 = match Self::db_get(&self.db, LAST_INSPECTED)? {
                     Some(n) => n + 1,
                     None => 0,
                 };
 
+                self.next_message_index
+                    .with_label_values(&[self.replica.name(), AGENT_NAME])
+                    .set(next_message_index as i64);
+
                 info!(
                     domain,
-                    nonce = next_to_inspect,
+                    nonce = next_message_index,
                     replica = self.replica.name(),
                     "Starting processor for {} {} at nonce {}",
                     domain,
                     self.replica.name(),
-                    next_to_inspect
+                    next_message_index
                 );
 
                 loop {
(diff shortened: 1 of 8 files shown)


rust/agents/processor/src/processor.rs
```diff
@@ -255,13 +261,23 @@ impl Processor {
         allowed: Option<HashSet<H256>>,
         denied: Option<HashSet<H256>>,
     ) -> Self {
+        let next_message_index = core
```
prestwich, line 264: i believe this is tracked per replica, as it's the next domain-specific nonce. Instead of having this one, we should dynamically register a new one per- `Replica` struct using the `new_replica_int_gauge` method?/

rust/chains/optics-ethereum/src/settings.rs
```diff
@@ -87,6 +87,66 @@ pub enum EthereumConnection {
 }
 
 impl EthereumConnection {
+    /// Watch network for new blocks and accumulate block height into `metric`
+    pub fn track_block_height(
```
emberian, line 91: @prestwich is this OK? i'm not sure this is the appropriate way to track this metric, but it's the way i found surfing the API docs.

rust/agents/processor/src/processor.rs
emberian: @prestwich one metric registered, each replica reports to it using its own name https://github.com/celo-org/optics-monorepo/pull/746/files#diff-9044bde722c9568d37c649b7c6a3366d225ade57ebdf4ab0dac8fe90062827acR101

prestwich: ahhhh interesting! I was unaware of how labels worked. This looks better than my kludge work :)

rust/chains/optics-ethereum/src/settings.rs
prestwich: this should work. the other option is to loop `provider.get_block_number` in a loop

This pattern would also work for other metrics. It seems reasonable to try to capture any ethereum-related metric we want here.

prestwich: can we do this in a way that generalizes to future connection types?
e.g. provide a standard trait or interface that captures this "register and track relevant metrics" behavior

rust/agents/processor/src/processor.rs
```diff
@@ -255,13 +261,23 @@ impl Processor {
...
+            .metrics
+            .new_int_gauge(
+                "next_message_index",
+                "Index of the next message to inspect",
+                &["replica", "agent"],
+            )
+            .expect("Processor metric alreadt registered -- should have be a singleton");
```
prestwich, line 271: typo `already`

prestwich (commented):
@yourbuddyconner we're going to want to run this on a 3-testnet environment to be sure it bheaves as expected

rust/agents/processor/src/processor.rs
yourbuddyconner: Labels are very cool indeed

rust/chains/optics-ethereum/src/settings.rs
emberian: from what i'm reading the state of the art is generally to use the global prometheus registry. can we implement at least one or two more before we try to generalize?

rust/agents/processor/src/processor.rs
```diff
@@ -294,25 +310,19 @@ impl OpticsAgent for Processor {
         let allowed = self.allowed.clone();
         let denied = self.denied.clone();
 
-        let metrics = self.core.metrics.clone();
+        let next_message_index = self.next_message_index.clone();
```
prestwich, line 313: it was not intuitive to me that prometheus metrics use an internal `Arc` and that cloning them preserves that arc. Can we document this somewhere?

rust/chains/optics-ethereum/src/settings.rs
prestwich: yeah, it's probably too early for something generic here.

prestwich (commented):
before merge, I would like to run agents on `dev` and ensure they work as intended

rust/agents/processor/src/processor.rs
emberian: it's a good thing, too! they carry around one of these:

```rust
pub struct Desc {
    /// fq_name has been built from Namespace, Subsystem, and Name.
    pub fq_name: String,
    /// help provides some helpful information about this metric.
    pub help: String,
    /// const_label_pairs contains precalculated DTO label pairs based on
    /// the constant labels.
    pub const_label_pairs: Vec<LabelPair>,
    /// variable_labels contains names of labels for which the metric
    /// maintains variable values.
    pub variable_labels: Vec<String>,
    /// id is a hash of the values of the ConstLabels and fqName. This
    /// must be unique among all registered descriptors and can therefore be
    /// used as an identifier of the descriptor.
    pub id: u64,
    /// dim_hash is a hash of the label names (preset and variable) and the
    /// Help string. Each Desc with the same fqName must have the same
    /// dimHash.
    pub dim_hash: u64,
}
```

the handle doesn't get cloned too often, and the extra per-instance storage isn't terribly sophisticated.

Screen Shot 2021-09-08 at 6 51 26 PM

that's the default behavior of `Arc` for any type, the usual promise you're supposed to make is that you "only use it for sharing". at least, doing otherwise starts to break `serde`'s `rc` feature, which makes for sadness. cheaper than your usual `.into()`!

yourbuddyconner:
Confirmed metrics *work* on dev environment! 

```
# HELP optics_block_height Height of a recently observed block
# TYPE optics_block_height gauge
optics_block_height{VERSION="0.1.0",agent="updater",network="alfajores"} 7189170
optics_block_height{VERSION="0.1.0",agent="updater",network="kovan"} 27139229
optics_block_height{VERSION="0.1.0",agent="updater",network="rinkeby"} 9269749
# HELP optics_signed_attestation_count Number of attestations signed
# TYPE optics_signed_attestation_count counter
optics_signed_attestation_count{VERSION="0.1.0",agent="updater",network="alfajores"} 6
```

```
# HELP optics_block_height Height of a recently observed block
# TYPE optics_block_height gauge
optics_block_height{VERSION="0.1.0",agent="processor",network="kovan"} 27139256
optics_block_height{VERSION="0.1.0",agent="processor",network="rinkeby"} 9269760
# HELP optics_next_message_index Index of the next message to inspect
# TYPE optics_next_message_index gauge
optics_next_message_index{VERSION="0.1.0",agent="processor",replica="alfajores"} 387
optics_next_message_index{VERSION="0.1.0",agent="processor",replica="kovan"} 387
```

yourbuddyconner (approved):

PR #751, opened 2021-09-08
Events: Process and Send

### LINKED ISSUE #754: Add `Send` event to Token Bridge
need to remove `Migrate` event in order to do this

### LINKED ISSUE #755: Unify `Process` events on Replica
make `ProcessSuccess` and `ProcessError` => `Process`

### LINKED ISSUE #756: Protect against too-long `returndata` in Replica
If return data is too long, copying it to memory may cause Replica process to run out of gas
we need to fix this

anna-carroll:
closes #754 
closes #755 
closes #756


solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
```diff
@@ -44,13 +44,20 @@ contract BridgeRouter is Version0, Router, TokenRegistry {
...
+    /**
+     * @notice emitted when tokens are sent from this domain to another domain
+     * @param token the address of the token contract
+     * @param from the address sending tokens
+     * @param toDomain the domain of the chain the tokens are being sent to
+     * @param toId the bytes32 address of the recipient of the tokens
+     * @param amount the amount of tokens sent
+     */
+    event Send(
+        address indexed token,
```
prestwich, line 56: should this emit the canonical token id instead?
Guess you can't index off that 🤔

solidity/optics-core/contracts/Replica.sol
```diff
@@ -64,23 +64,15 @@ contract Replica is Version0, Common {
...
+     * @param returnData the return data from the external call
      */
-    event ProcessError(
+    event Process(
         bytes32 indexed messageHash,
-        uint32 indexed nonce,
-        address indexed recipient,
-        bytes returnData
+        bool indexed success,
+        bytes indexed returnData
```
prestwich, line 75: Thought:
- a contract can return an extremely large amount of returndata and cause this to out-of-gas
- this would cause some annoyance wrt the processor, but not the end of the world

prestwich: 1 more reason we can never guarantee processing order I guess 😬

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
anna-carroll: mmm good question, i was thinking about including a boolean `localOrigin` which is true if it's a local token - but this would make more sense 🤔

anna-carroll: > Guess you can't index off that
could you elaborate? why not

solidity/optics-core/contracts/Replica.sol
anna-carroll: ticket to track: #756 
@prestwich going to push a fix

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
prestwich: well, you can, but indexing off a struct is a bit of a pain right? haven't tried it in ages

prestwich (approved):

solidity/optics-core/contracts/Replica.sol
prestwich: #757

solidity/optics-xapps/contracts/bridge/BridgeRouter.sol
anna-carroll: ah i thought you meant the packed bytes `tokenId`

PR #757, opened 2021-09-08
defense: prevent copying long returndata in replica.process

prestwich:
tests pass locally :)

this PR changes the call semantics in ` replica.process` to use assembly to access only up to 256 bytes of returndata, rather than copying all of it to memory


prestwich:
closes #756

solidity/optics-core/contracts/Replica.sol
```diff
@@ -187,17 +187,41 @@ contract Replica is Version0, Common {
...
-                "handle(uint32,bytes32,bytes)",
-                _m.origin(),
-                _m.sender(),
-                _m.body().clone()
+        // by assembly calling "handle" function
+        // we call via assembly to avoid memcopying a very large returndata
+        // returned by a malicious contract
+        assembly {
+            _success := call(
+                _gas, // gas
```
anna-carroll, line 207: can you not use `PROCESS_GAS` in the assembly?

solidity/optics-core/contracts/Replica.sol
```diff
@@ -187,17 +187,41 @@ contract Replica is Version0, Common {
         require(gasleft() >= PROCESS_GAS + RESERVE_GAS, "!gas");
         // get the message recipient
         address _recipient = _m.recipientAddress();
+        // set up for assembly call
+        uint256 _toCopy;
```
anna-carroll, line 191: nit: rename `_returnDataLength` or something like that

solidity/optics-core/contracts/Replica.sol
```diff
@@ -187,17 +187,41 @@ contract Replica is Version0, Common {
...
+        uint256 _gas = PROCESS_GAS;
+        // allocate memory for returndata
+        bytes memory _returnData = new bytes(0x100);
```
anna-carroll, line 194: `0x100` is magic number repeated several times, would it be worth storing as `_maxReturnDataLength`

solidity/optics-core/contracts/Replica.sol
```diff
@@ -187,17 +187,41 @@ contract Replica is Version0, Common {
...
-                _m.body().clone()
+        // by assembly calling "handle" function
+        // we call via assembly to avoid memcopying a very large returndata
+        // returned by a malicious contract
+        assembly {
+            _success := call(
+                _gas, // gas
+                _recipient, // recipient
+                0, // ether value
+                add(_calldata, 0x20), // inloc
```
anna-carroll, line 210: for my own learning: what is `0x20` here

solidity/optics-core/contracts/Replica.sol
prestwich: constants aren't available in assembly, iirc because they are actually shortcuts for `codecopy`

solidity/optics-core/contracts/Replica.sol
prestwich: I think `_toCopy` is more accurate, as it is the number of bytes we copy, which may be less than the returndatalength

solidity/optics-core/contracts/Replica.sol
prestwich: 32 bytes, the length of 1 EVM word

solidity/optics-core/contracts/Replica.sol
prestwich: roger wilco

anna-carroll (approved):

PR #774, opened 2021-09-10
refactor: improve db and test roundtrip store/load

prestwich:


yourbuddyconner (approved):
Tested in prod, works! :D

PR #776, opened 2021-09-11
ship spans to honeycomb

emberian:
seems like a useful tool, and their telemetry layer seems easier to use than opentelemetry....


yourbuddyconner:
I would like to see a discussion thread about this, with examples of the free-tier. Since it is free I don't really have an issue here but it isn't compatible with our current Grafana setup.

PR #781, opened 2021-09-13
Prover sync refactoring

prestwich:
Behavior is correct locally

Note that this disables the Watcher in Cargo.toml, pending #782


yourbuddyconner (approved):
yeet

PR #783, opened 2021-09-14
Prestiwch/polygon metrics fix

prestwich:
- (temporarily) disables replica height tracking
- moves height tracking to HomeIndexer task


rust/agents/kathy/src/kathy.rs
```diff
@@ -38,13 +38,15 @@ impl Kathy {
 
 #[async_trait]
 impl OpticsAgent for Kathy {
+    const AGENT_NAME: &'static str = "kathy";
```
emberian, line 41: ooooooo associated constants! i forgot about those. this is a lot nicer than trying to jam structure into rust modules, which are unstructured bags.

rust/agents/kathy/src/kathy.rs
```diff
@@ -38,13 +38,15 @@ impl Kathy {
...
+
     type Settings = Settings;
 
     async fn from_settings(settings: Settings) -> Result<Self> {
         Ok(Self::new(
             settings.interval.parse().expect("invalid u64"),
             settings.chat.into(),
-            settings.base.try_into_core("kathy").await?,
+            settings.base.try_into_core(Self::AGENT_NAME).await?,
```
emberian, line 49: random thought: might be nice to have something like an `fn AgentCore::wrap(inner: impl OpticsAgent) -> Self`. the interplay between the SDK and the agents was a source of confusion for me.

rust/agents/relayer/src/relayer.rs
```diff
@@ -116,6 +118,8 @@ impl Relayer {
 #[async_trait]
 #[allow(clippy::unit_arg)]
 impl OpticsAgent for Relayer {
+    const AGENT_NAME: &'static str = AGENT_NAME;
```
emberian, line 121: beta-reduce this pls (remove the out-of-line defn)

rust/agents/updater/src/main.rs
```diff
@@ -27,9 +27,25 @@ async fn _main() -> Result<()> {
...
+    let block_height = agent
+        .as_ref()
+        .metrics
+        .new_int_gauge(
+            "block_height",
+            "Height of a recently observed block",
+            &["network", "agent"],
+        )
+        .expect("failed to register block_height metric")
+        .with_label_values(&[agent.home().name(), Updater::AGENT_NAME]);
```
emberian, line 44: aaaaah! what _is_ this type! good idea. i wonder if it caches any computation. might want to hoist `with_label_values` out of other loops if it does.

rust/agents/updater/src/updater.rs
```diff
@@ -248,6 +248,8 @@ impl Updater {
 // Ideally this hould be generic across all signers.
 // Right now we only have one
 impl OpticsAgent for Updater {
+    const AGENT_NAME: &'static str = AGENT_NAME;
```
emberian, line 251: beta-reduce

emberian: (can refer to the constant as `Updater::AGENT_NAME` elsewhere)

rust/agents/watcher/src/watcher.rs
```diff
@@ -309,6 +311,8 @@ impl Watcher {
 #[async_trait]
 #[allow(clippy::unit_arg)]
 impl OpticsAgent for Watcher {
+    const AGENT_NAME: &'static str = AGENT_NAME;
```
emberian, line 314: beta-reduce

rust/optics-base/src/agent.rs
```diff
@@ -106,9 +112,21 @@ pub trait OpticsAgent: Send + Sync + std::fmt::Debug + AsRef<AgentCore> {
     #[allow(clippy::unit_arg, unused_must_use)]
     #[tracing::instrument(err)]
     async fn run_all(&self) -> Result<()> {
+        let block_height = self
+            .as_ref()
+            .metrics
+            .new_int_gauge(
+                "block_height",
+                "Height of a recently observed block",
```
emberian, line 120: do you have insight into which blocks actually make it into this agent? the data flows from the chain -> ethers-rs -> everywhere else is opaque to me.

emberian (changes requested):
does this mean that there are ethereum RPCs that not all ethereum impls implement? what does `Ethereum Compatible` actually mean?

prestwich:
> does this mean that there are ethereum RPCs that not all ethereum impls implement? what does `Ethereum Compatible` actually mean?

the RPC has had several versions over the years with multiple slightly-incompatible implementations of each version. and then there are multiple API providers who implement custom subsets. It's a real mess

So yeah, in this case the Polygon RPC does not support the block filter call 😬

rust/agents/kathy/src/kathy.rs
prestwich: tbh I forgot about them til just now lmao

emberian:
Screen Shot 2564-09-14 at 15 36 59

someone call the internet police or at least the IETF

emberian (approved):

rust/optics-base/src/agent.rs
prestwich: sometime let's do an info session on this

rust/agents/updater/src/main.rs
prestwich: The updater main is unfortunately messy, but performs the same function as `OpticsAgent::run_all`

unrelated, should we consider pulling the prometheus `Registry` out of the agent core into a global `once_cell` or similar?

yourbuddyconner (approved):
yeet

yourbuddyconner (approved):

PR #791, opened 2021-09-15
bug: ProverSync from_disk now properly recovers tree state, proofs, etc.

prestwich:
- There was a significant bug where `from_disk` would cause `incremental` and `prover` to have different states
- while fixing this, I did some significant refactors to prover sync to have it rely on HomeIndexer and not block
- also made it run a proof-presence check on startup to ensure that we don't somehow miss proofs

- To consider: ProverSync no longer makes RPC requests so it is 100% sync now. What's a reasonable way (besides `wait_for_leaf`) to make it yield priority to other tasks?

### DIFF
--- a/rust/agents/processor/src/prover_sync.rs
+++ b/rust/agents/processor/src/prover_sync.rs
@@ -68,15 +68,22 @@ pub enum ProverSyncError {
 
 impl ProverSync {
     fn store_proof(&self, leaf_index: u32) -> Result<(), ProverSyncError> {
-        let proof = self.prover.prove(leaf_index as usize)?;
-        self.db.store_proof(leaf_index, &proof)?;
-        info!(
-            leaf_index,
-            root = ?self.prover.root(),
-            "Storing proof for leaf {}",
-            leaf_index
-        );
-        Ok(())
+        match self.prover.prove(leaf_index as usize) {
+            Ok(proof) => {
+                self.db.store_proof(leaf_index, &proof)?;
+                info!(
+                    leaf_index,
+                    root = ?self.prover.root(),
+                    "Storing proof for leaf {}",
+                    leaf_index
+                );
+                Ok(())
+            }
+            // ignore the storage request if it's out of range
+            Err(ProverError::ZeroProof { index: _, count: _ }) => Ok(()),
+            // bubble up any other errors
+            Err(e) => Err(e.into()),
+        }
     }
 
     /// Given rocksdb handle `db` containing merkle tree leaves,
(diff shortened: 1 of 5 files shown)


emberian:
> To consider: ProverSync no longer makes RPC requests so it is 100% sync now. What's a reasonable way (besides wait_for_leaf) to make it yield priority to other tasks?

don't! task scheduling is the responsibility of the execution context. use `tokio::task::spawn_blocking`, and it will take care of the rest.

rust/optics-core/src/db/mod.rs
```diff
@@ -257,32 +307,20 @@ impl DB {
...
-        let mut path = Path::new(".").canonicalize()?;
-        path.extend(&[db_path]);
-
-        match path.is_dir() {
-            true => info!(
-                "Opening existing db at {path}",
-                path = path.to_str().unwrap()
-            ),
-            false => info!("Creating db at {path}", path = path.to_str().unwrap()),
+    // TODO(james): this is a quick-fix for the prover_sync and I don't like it
```
emberian, line 313: eventually consistent, inside the same process. i dig it.

rust/agents/processor/src/prover_sync.rs
```diff
@@ -69,14 +67,67 @@ pub enum ProverSyncError {
...
-    pub fn new(prover: Prover, home: Arc<Homes>, db: DB) -> Self {
-        Self {
+    fn store_proof(&self, leaf_index: u32) -> Result<(), ProverSyncError> {
+        let proof = self.prover.prove(leaf_index as usize)?;
+        self.db.store_proof(leaf_index, &proof)?;
+        info!(
+            leaf_index,
+            root = ?self.prover.root(),
+            "Storing proof for leaf {}",
+            leaf_index
```
emberian, line 77: how do you feel about `"Storing proof", leaf_index = leaf_index` ? although i don't know that the short formatter includes fields...

rust/agents/processor/src/prover_sync.rs
```diff
@@ -69,14 +67,67 @@ pub enum ProverSyncError {
...
+        let sync = Self {
             prover,
-            home,
-            incremental: IncrementalMerkle::default(),
+            incremental,
             db,
+        };
+
+        // Ensure proofs exist for all leaves
+        for i in 0.. {
```
emberian, line 119: 👍🏼 i love executable postconditions

emberian (approved):

rust/agents/processor/src/prover_sync.rs
prestwich: ah you mean using named formatter args in the format string? I don't see a compelling reason to do it but I don't mind it

rust/agents/processor/src/prover_sync.rs
prestwich: this made me think of an edge case I want to handle. will push up a commit in a sec

emberian (approved):

PR #793, opened 2021-09-15
Balance exporter tool

emberian:
Still to do:

  - [ ] query the chain
  - [ ] export the metrics

### DIFF
--- a/rust/balance-exporter/src/main.rs
+++ b/rust/balance-exporter/src/main.rs
@@ -1,0 +1,160 @@
+use std::{path::PathBuf, time::Duration};
+
+use clap::Arg;
+use color_eyre::eyre::anyhow;
+use futures::StreamExt;
+use human_panic::setup_panic;
+use optics_base::settings::ChainSetup;
+use tokio::time::Instant;
+
+#[derive(serde::Deserialize, Debug)]
+struct Input {
+    contracts: Vec<ChainSetup>,
+}
+
+struct Sample {
+    balances: Vec<color_eyre::Result<String>>,
+}
+
+async fn poll_once(input: &Input, timeout: Duration) -> Sample {
+    // does this open a new ws connection for each query? probably.
+    let (send, mut recv) = tokio::sync::mpsc::unbounded_channel();
+
+    // moves `send` into the subtask executor. by the time it is complete,
+    // `recv` will complete to exhaustion, due to `send` clones being dropped
+    futures::stream::iter(
+        input
+            .contracts
+            .iter()
+            .enumerate()
+            .zip(std::iter::repeat(send)),
+    )
+    .for_each_concurrent(None, |((ix, _cs), send)| async move {
+        let sub_result =
+            tokio::time::timeout(timeout, std::future::ready(Err(anyhow!("WIP")))).await;
+        send.send((ix, sub_result))
+            .expect("channel closed before we're done?");
+    })
+    .await;
+
+    let mut balances: Vec<Option<color_eyre::Result<String>>> =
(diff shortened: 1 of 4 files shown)


rust/balance-exporter/src/main.rs
```diff
@@ -0,0 +1,160 @@
...
+    futures::stream::iter(
+        input
+            .contracts
+            .iter()
+            .enumerate()
+            .zip(std::iter::repeat(send)),
+    )
+    .for_each_concurrent(None, |((ix, _cs), send)| async move {
+        let sub_result =
+            tokio::time::timeout(timeout, std::future::ready(Err(anyhow!("WIP")))).await;
```
emberian, line 34: it's unreasonable to create a new `ethers` `Provider` for each individual `ChainSetup` we come across. perhaps something like `Map<String, Box<dyn ChainApi>>`, mapping `name` fields to our abstraction of what a `chain` is?

at startup, run through the list of setups and for each network not present in the map, cache a new `Provider` connection? `optics_base` divorcing live connections from connection-creation-metadata is a bit frustrating here.

prestwich: > optics_base divorcing live connections from connection-creation-metadata is a bit frustrating here

agreed. I think keeping a 

> perhaps something like Map<String, Box<dyn ChainApi>>, mapping name fields to our abstraction of what a chain is?

seems reasonable. go ahead and do this. create the trait, but make it as minimal as possible, as the union of chain connections methods is likely to be small :*

rust/agents/processor/src/processor.rs
```diff
@@ -135,7 +128,7 @@ impl Replica {
     /// In case of error: send help?
     #[instrument(err, skip(self), fields(self = %self))]
     async fn try_msg_by_domain_and_nonce(&self, domain: u32, nonce: u32) -> Result<bool> {
-        use optics_core::traits::Replica;
+        use optics_core::Replica;
```
prestwich, line 131: why not at top of file?

rust/agents/watcher/src/watcher.rs
```diff
@@ -794,7 +794,7 @@ mod test {
                 db: db,
                 indexer: IndexSettings::default(),
                 metrics: Arc::new(
-                    optics_base::metrics::CoreMetrics::new(
+                    optics_base::new(
```
prestwich, line 797: not sure I like `module::new` here 🤔

emberian: whoa that one is pretty whack, i wonder what is re-exporting `CoreMetrics::*`...

rust/agents/processor/src/processor.rs
emberian: /me hoists

rust/agents/watcher/src/watcher.rs
emberian: nothing? does this even compile? confusing.

emberian: aha, it's not included in `cargo test`:

```
error: current package believes it's in a workspace when it's not:
current:   /Users/ember/optics-monorepo/rust/agents/watcher/Cargo.toml
workspace: /Users/ember/optics-monorepo/rust/Cargo.toml

this may be fixable by adding `agents/watcher` to the `workspace.members` array of the manifest located at: /Users/ember/optics-monorepo/rust/Cargo.toml
Alternatively, to keep it out of the workspace, add the package to the `workspace.exclude` array, or add an empty `[workspace]` table to the package's manifest.
```

PR #801, opened 2021-09-17
feature: dedup DB keys

ltchang2019:
- all db methods take home_name as added prefix
- lots of repeated code given that every db method requires a home_name (consider pairing DB with home name?)

### DIFF
--- a/rust/optics-core/src/db/mod.rs
+++ b/rust/optics-core/src/db/mod.rs
@@ -15,18 +15,20 @@ use crate::{
 
 use self::iterator::PrefixIterator;
 
-static NONCE: &str = "destination_and_nonce_";
-static LEAF_IDX: &str = "leaf_index_";
-static LEAF_HASH: &str = "leaf_hash_";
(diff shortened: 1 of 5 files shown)


rust/optics-core/src/db/mod.rs
```diff
@@ -15,18 +15,20 @@ use crate::{
...
+static LATEST_LEAF: &str = "_latest_known_leaf_";
 
-static LATEST_LEAF: &str = "latest_known_leaf_";
-
-#[derive(Debug, Clone)]
 /// A KV Store
+///
+/// Key structure: ```<home_name>_<type_prefix>_<key>```
+#[derive(Debug, Clone)]
 pub struct DB(Arc<Rocks>);
```
emberian, line 32: upgrade this from a handle to a struct with the `home_name`? are there places where `home_name` isn't present?

emberian (changes requested):

ltchang2019:
Closing in favor of #803

PR #803, opened 2021-09-17
refactor: deduplicate db keys by adding `HomeDB`

ltchang2019:
- adds wrapper struct `HomeDB`, which contains `home_name` and all home-specific db operations
- `updater`, `processor`, and `proversync` sub-components now all use `HomeDB` while top-level agents themselves retain handle to plain `db` (stored in `AgentCore`); allows agents access to db if needed in future but all current operations store home-specific data
- CONSIDER: if agent subcomponents really need to access raw DB in manner not related to specific home, we can make `HomeDB.db` public OR the subcomponents could contain both DB and HomeDB

Closes #801


rust/optics-core/src/db/home_db.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    fn full_prefix(&self, prefix: impl AsRef<[u8]>) -> Vec<u8> {
+        let mut full_prefix = vec![];
+        full_prefix.extend(self.home_name.as_ref() as &[u8]);
+        full_prefix.extend(prefix.as_ref());
+        full_prefix
+    }
+
+    /// Store encodable with `self.home_name` prefixed to key
```
emberian, line 47: Implementation details leaked into the docs. In general these docs aren't helpful - I'd rather omit them and #![allow(missing_docs)] than simply repeat the signature.

rust/optics-core/src/db/mod.rs
```diff
@@ -1,29 +1,16 @@
...
 
 /// Shared functionality surrounding use of rocksdb
 pub mod iterator;
 
-use crate::{
-    accumulator::merkle::Proof, traits::RawCommittedMessage, utils, Decode, Encode, OpticsError,
-    OpticsMessage, SignedUpdate,
-};
+/// Home-specific db operations
+pub mod home_db;
```
emberian, line 10: keep the module private and re-export all of its public members. having both makes the docs unreadable.

rust/optics-core/src/db/mod.rs
```diff
@@ -1,29 +1,16 @@
...
-static PREV_ROOT: &str = "update_prev_root_";
-static NEW_ROOT: &str = "update_new_root_";
-static LATEST_ROOT: &str = "update_latest_root_";
-static PROOF: &str = "proof_";
-
-static LATEST_LEAF: &str = "latest_known_leaf_";
+use crate::{Decode, Encode, OpticsError};
 
 #[derive(Debug, Clone)]
 /// A KV Store
```
emberian, line 16: this could stand to be renamed to `TypedDB` or something like that.

rust/optics-core/src/db/home_db.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub fn store_leaf(
+        &self,
+        leaf_index: u32,
+        destination_and_nonce: u64,
+        leaf_hash: H256,
+    ) -> Result<(), DbError> {
+        debug!(
+            leaf_index,
+            leaf_hash = ?leaf_hash,
+            "storing leaf hash keyed by index and dest+nonce"
```
emberian, line 135: at the least, the debug message and the docs on the method should agree -- they're communicating the same thing to different people.

rust/optics-core/src/db/home_db.rs
```diff
@@ -0,0 +1,273 @@
...
+    pub fn store_update(&self, update: &SignedUpdate) -> Result<(), DbError> {
+        debug!(
+            previous_root = ?update.update.previous_root,
+            new_root = ?update.update.new_root,
+            "storing update in DB"
+        );
+
+        // If there is no latet root, or if this update is on the latest root
+        // update latest root
+        match self.retrieve_latest_root()? {
```
emberian, line 206: `if Some(update.update.previous_root) == self.retrieve_latest_root()?`

rust/optics-core/src/db/home_db.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    fn store_latest_root(&self, root: H256) -> Result<(), DbError> {
+        debug!(root = ?root, "storing new latest root in DB");
+        self.store_encodable("", LATEST_ROOT, &root)
+    }
+
+    /// Store a signed update
+    pub fn store_update(&self, update: &SignedUpdate) -> Result<(), DbError> {
+        debug!(
```
emberian, line 198: I'd prefer if all of the events were instead spans. that way they carry timing. the storage isn't instantaneous - it carries on until the method is complete. it also makes network and flash storage behavior trace the same.

emberian (dismissed):
the refactor looks good, but now that I've read the storage code I have questions :)

rust/optics-core/src/db/mod.rs
ltchang2019: yeah good point. refactoring this to split up `TypedDB` and `HomeDB`

rust/optics-core/src/db/home_db.rs
ltchang2019: #813

rust/optics-core/src/db/home_db.rs
```diff
@@ -0,0 +1,273 @@
...
+            "storing update in DB"
+        );
+
+        // If there is no latet root, or if this update is on the latest root
+        // update latest root
+        match self.retrieve_latest_root()? {
+            Some(root) => {
+                if root == update.update.previous_root {
+                    self.store_latest_root(update.update.new_root)?;
+                }
```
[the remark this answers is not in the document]
ltchang2019: `store_update` called by updater, watcher, and home_indexer:
- updater hits this path: updater skipped storing an update in its db (its db is missing 1+ roots between the root from the current update and its last stored root)
- watcher hits this path: watcher's `UpdateHandler` has polled an already-observed root from the home_indexer, which means that the home_indexer is starting from the wrong block height and is sending back already-seen data
- home_indexer hits this path: same as above case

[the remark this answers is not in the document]
ltchang2019: For indexing issues, it could also be that updates are _not_ in order

emberian (approved):

PR #809, opened 2021-09-20
refactor: watcher home indexing rewrite

ltchang2019:
- watcher now employs home indexing logic
- `AgentCore::run_many` is NOT used anymore (`AgentCore::run` was already not used)
  - `run_many` can't be used because it borrows `&self`, making it impossible to run home indexing task concurrently with run_loop
- run_loop does NOT return `DoubleUpdate` anymore because in order to use `select_all` to race the indexing and run loop tasks, the return types of the tasks must be the same
  - run_loop instead uses a tokio::oneshot to transmit the potential `double_update`


rust/agents/watcher/src/watcher.rs
```diff
@@ -357,67 +423,83 @@ impl OpticsAgent for Watcher {
...
+            for task in remaining.into_iter() {
+                cancel_task!(task);
+            }
+            self.shutdown().await;
+
+            // Check if double update was sent during run task
+            match double_update_rx.try_recv() {
+                Ok(double_update) => {
+                    tracing::error!(
+                        "Double update detected! Notifying all contracts and unenrolling replicas!"
```
emberian, line 483: please include some of the information from `double_update` in this event so that event consumers (eg, us later) have some correlation keys.

rust/agents/watcher/src/watcher.rs
```diff
@@ -357,67 +423,83 @@ impl OpticsAgent for Watcher {
...
-            ContractWatcher::new(self.interval_seconds, from, tx.clone(), home.clone())
-                .spawn()
-                .in_current_span();
-        let home_sync = HistorySync::new(self.interval_seconds, from, tx.clone(), home)
-            .spawn()
-            .in_current_span();
-
-        let join_result = handler.await;
+    fn run_many(&self, _replicas: &[&str]) -> Instrumented<JoinHandle<Result<()>>> {
+        tokio::spawn(
```
emberian, line 427: can you replace this with a simple `panic!`? this approach is cute but needless

rust/agents/watcher/src/watcher.rs
ltchang2019: Added the double_update to `tracing::error!`. Lmk if that's what you meant

prestwich:
have you looked at how `run_all` is laid out in the processor? it solves the same problem of borrowed self while still running multiple tasks

PR #823, opened 2021-09-22
Prestwich/prove cli

prestwich:
@yourbuddyconner this tool allows you to prove individual leaves from a DB dump.

please ensure that changing the kms-cli path does not break your workflow


rust/tools/prove-cli/README.md
```diff
@@ -0,0 +1,21 @@
...
+- `cargo run --bin prove-cli`
+  - `--leaf-index` OR `--leaf_hash` - specify the leaf to prove
+    - `leaf_index` takes precedence if both are specified
+  - `--rpc` specify the RPC endpoint
+  - `--key` specify the hex key to use to sign txns
+    - in future versions this will be an env var or a node or aws signer
+  - `--db` specify the filepath to the DB
+  - `--address` specify the Replica address to submit to
+
+### Exampl
```
emberian, line 17:
```suggestion
### Example
```

rust/tools/prove-cli/README.md
```diff
@@ -0,0 +1,21 @@
+## Prover CLI
+
+This CLI directly accesses a synced or partially-synced processor DB, retrieves
+messages and proofs, and dispatches them to the contracts.
+
+### Usage
+
+- `cargo run --bin prove-cli`
+  - `--leaf-index` OR `--leaf_hash` - specify the leaf to prove
+    - `leaf_index` takes precedence if both are specified
```
emberian, line 10: discrepancy

rust/tools/prove-cli/src/main.rs
```diff
@@ -0,0 +1,107 @@
...
+
+#[derive(Clap)]
+struct Opts {
+    /// Leaf index to prove
+    #[clap(short, long)]
+    leaf_index: Option<u32>,
+
+    /// Leaf index to prove
+    #[clap(short, long)]
+    leaf_hash: Option<H256>,
```
emberian, line 31:
```
thread 'main' panicked at 'Short option names must be unique for each argument, but '-l' is in use by both 'leaf-index' and 'leaf-hash'', /Users/e/.cargo/registry/src/github.com-1ecc6299db9ec823/clap-3.0.0-beta.4/src/build/app/debug_asserts.rs:105:17
```

emberian: (just remove the `short`s)

emberian (dismissed):

ltchang2019 (approved):
Didn't manually run the cli tool (maybe @yourbuddyconner should before merge) but logic-wise LGTM

yourbuddyconner:
As mentioned in DMs, I would love it if this tool: 
- Stored the default replicas for whatever home we're processing messages for
- Allowed me to configure all three RPC endpoints and would just pick the right one (env vars fine, config file also fine) 

Something like: 
`prove-cli --home celo --leaf_index 42 --key <KEY> --db` 

I would make sure `--key` is funded on all three networks so it would generally just work. 

The CLI would discover which replica it needs to process the message on, and then use the right RPC endpoint.

prestwich:
> As mentioned in DMs, I would love it if this tool:
> 
> * Stored the default replicas for whatever home we're processing messages for
> * Allowed me to configure all three RPC endpoints and would just pick the right one (env vars fine, config file also fine)
> 
> Something like:
> `prove-cli --home celo --leaf_index 42 --key <KEY> --db ~/path/to/db`
> 
> I would make sure `--key` is funded on all three networks so it would generally just work.
> 
> The CLI would discover which replica it needs to process the message on, and then use the right RPC endpoint.

wish granted

yourbuddyconner (approved):
Ship it

Issue #812, opened 2021-09-20
bug: processor crash - ProverError(ZeroProof)


yourbuddyconner:
```
{"timestamp":"Sep 20 18:15:57.064","level":"INFO","fields":{"message":"Reloaded ProverSync from disk","target_latest_root":"0xf7104deced1063f0bc0dc70be10abddfe3f92f34dcc1fe8d860bfcea83ec433d","root":"0xf7104deced1063f0bc0dc70be10abddfe3f92f34dcc1fe8d860bfcea83ec433d"},"target":"processor::prover_sync"}
The application panicked (crashed).
Message: db error: ProverError(ZeroProof { index: 203, count: 203 })
```

Additional investigation required.

ltchang2019:
Crash happening here: https://github.com/celo-org/optics-monorepo/blob/main/rust/agents/processor/src/prover_sync.rs#L131

Looks like there is a leaf at `index` where `index == total_num_of_leaves`... looking into this more

ltchang2019:
Crash happening here:
- https://github.com/celo-org/optics-monorepo/blob/main/rust/agents/processor/src/prover_sync.rs#L131 (top level)
- https://github.com/celo-org/optics-monorepo/blob/main/rust/agents/processor/src/prover_sync.rs#L71 (next level in call stack)

Looks like there is a leaf at `index` where `index == total_num_leaves_ingested` which should not be the case... still looking

yourbuddyconner:
More details, this just occured on `Celo Home Processor`: 

```
{"timestamp":"Sep 21 19:20:20.148","level":"INFO","fields":{"message":"Opening existing db at /usr/share/optics"},"target":"optics_core::db","span":{"db_path":"\"/usr/share/optics\"","name":"from_path"},"spans":[{"db_path":"\"/usr/share/optics\"","name":"from_path"}]}
{"timestamp":"Sep 21 19:20:21.013","level":"INFO","fields":{"message":"starting prometheus server on 0.0.0.0:9090","port":9090},"target":"optics_base::metrics"}
{"timestamp":"Sep 21 19:20:21.013","level":"INFO","fields":{"message":"listening on http://0.0.0.0:9090"},"target":"warp::server","span":{"addr":"0.0.0.0:9090","name":"Server::run"},"spans":[]}
{"timestamp":"Sep 21 19:20:21.013","level":"INFO","fields":{"message":"Starting Processor tasks"},"target":"processor::processor"}
[...]

PR #824, opened 2021-09-22
fix: fix potential out-of-order updates when `HomeIndexer` syncing

ltchang2019:
- Orders queried update events by `blocknumber` and `transaction_index` within block (should solve skipped messages issue?)
- Stop proversync leaf/proof check at `prover.count()` (don't try to build proofs for leaves without an update yet)
- Adds `warn!` when we attempt to store root that doesn't build off latest root (shouldn't happen anymore though after fix)

Potentially related to #812

### DIFF
--- a/rust/agents/processor/src/prover_sync.rs
+++ b/rust/agents/processor/src/prover_sync.rs
@@ -79,7 +79,8 @@ impl ProverSync {
                 );
                 Ok(())
             }
-            // ignore the storage request if it's out of range
+            // ignore the storage request if it's out of range (e.g. leaves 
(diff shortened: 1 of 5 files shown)


rust/agents/processor/src/prover_sync.rs
```diff
@@ -123,7 +124,7 @@ impl ProverSync {
         };
 
         // Ensure proofs exist for all leaves
-        for i in 0.. {
+        for i in 0..sync.prover.count() as u32 {
```
prestwich, line 127: 👍

prestwich (approved):
LGTM

PR #839, opened 2021-09-24
feature: optionally skip replica tasks

prestwich:
Adds a `sync` config param. Setting this param to any string value will put the processor in sync-only mode. The processor will not dispatch txns to replicas, it will only sync the DB.

### DIFF
--- a/rust/agents/processor/src/settings.rs
+++ b/rust/agents/processor/src/settings.rs
@@ -11,4 +11,6 @@ decl_settings!(Processor {
     allowed: Option<HashSet<H256>>,
     /// A deny list of message senders
     denied: Option<HashSet<H256>>,
+    /// Only index transactions if this key is set
+    index: Option<String>,
 });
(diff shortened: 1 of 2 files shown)


rust/agents/processor/src/settings.rs
```diff
@@ -11,4 +11,6 @@ decl_settings!(Processor {
     allowed: Option<HashSet<H256>>,
     /// A deny list of message senders
     denied: Option<HashSet<H256>>,
+    /// Only index transactions if this key is set
+    index: Option<String>,
```
ltchang2019, line 15: It looks like if `index` is set, then the processor will only run the `index` and `sync` tasks but will not process txs on replicas. It's a little confusing the way `index.is_none()` equates to `process == true`. Maybe have these bools not be negations of each other and keep them on same side? 

Like instead maybe `settings.only_sync.is_some()` equates to `processor.only_sync == true` and we do
```
if !self.only_sync {
  ...run_many processing tasks...
}
```

ltchang2019 (approved):

rust/agents/processor/src/settings.rs
prestwich: because of limitations in the settings crate, we want to do 1-word settings in all cases. Do you have a more-clear 1 word setting idea?

ltchang2019 (approved):

PR #802, opened 2021-09-17
tool: Contract Event Metrics

yourbuddyconner:

PR #841, opened 2021-09-24
Print unprocessed messages to file

anna-carroll:
At the moment this is *very* hacky and should not be merged. 
`git checkout anna/monitor && cd typescript/optics-monitor && npm run monitor` to print un-processed messages from each chain to files


anna-carroll:
**NOTE:** this does NOT check how old the unprocessed messages are. all messages that have not yet been processed will show up. some of these messages may just be recent and still awaiting the challenge period.

anna-carroll:
moved code to #802 PR in this commit: https://github.com/celo-org/optics-monorepo/pull/802/commits/ac90d7a28dbafed4b250fbece7df2e1fbc4a6530

Issue #827, opened 2021-09-23
fix: OpticsMessage class breaks if multiple messages in same tx


anna-carroll:
OpticsMessage class in the Multi-Provider takes a transaction receipt in its constructor. It looks for the first Dispatch message within the transaction, and is instantiated with that. 

this doesn't work if multiple messages are dispatched in one transaction.

the class should probably intake a Dispatch log instead of a whole transaction receipt

Issue #815, opened 2021-09-21
improve OpticsMessage status in multi-provider


anna-carroll:
right now, the multi-provider only returns the messages status on the Replica contract 

we should have a way to know the message status at a system-wide level: https://github.com/celo-org/optics-monorepo/issues/641#issuecomment-915173723

PR #842, opened 2021-09-24
feat/refactor: Multi-Provider improvements

### LINKED ISSUE #835: Add .events() to OpticsMessage class
Query for array of events updating the status of a message

**Event type**
- timestamp 
- transaction hash 
- chain domain 
- chain name 
- block number 
- event details - log / topics / parsed / etc 

**Events**
- Dispatched
- Updated on Home
- Updated on Replica
- Processed

anna-carroll:
Closes #835 #827 #815


prestwich:
Initially it was instantiated from an event. The reason I changed it to be instantiated from a receipt was to preserve blockheight, and other tx confirmation info. Does this PR preserve those?

anna-carroll:
TransactionReceipt (which is ContractReceipt without parsed Events array):
```
export interface TransactionReceipt {
    to: string;
    from: string;
    contractAddress: string,
    transactionIndex: number,
    root?: string,
    gasUsed: BigNumber,
    logsBloom: string,
    blockHash: string,
    transactionHash: string,
    logs: Array<Log>,
    blockNumber: number,
    confirmations: number,
    cumulativeGasUsed: BigNumber,
    effectiveGasPrice: BigNumber,
    byzantium: boolean,
    type: number;
    status?: number
};
```
Log (which we can swap out for Event here):
```
export interface Log {
    blockNumber: number;
    blockHash: string;
    transactionIndex: number;

    removed: boolean;

    address: string;
    data: string;

    topics: Array<string>;

    transactionHash: string;
    logIndex: number;
}
```

diff (which we can get by querying the transaction if we want)
```
    to: string;
    from: string;
    transactionIndex: number,
    root?: string,
    gasUsed: BigNumber,
    logsBloom: string,
    confirmations: number,
    cumulativeGasUsed: BigNumber,
    effectiveGasPrice: BigNumber,
    byzantium: boolean,
    type: number;
    status?: number
```

anna-carroll:
you know what, i will just pass in the event and the contract receipt

typescript/optics-provider/src/provider.ts
```diff
@@ -106,4 +115,9 @@ export class MultiProvider {
 
     return await signer?.getAddress();
   }
+
+  async getTransactionReceipt(nameOrDomain: string | number, transactionHash: string): Promise<TransactionReceipt> {
```
prestwich, line 119: design question: how many standard provider methods should we be wrapping and re-surfacing, as opposed to making the use fetch the provider?


e.g. what is preferred

```
// more boilerplate, less verbose for user
multi.getTransactionReceipt('ethereum', '0x...')

/// OR 

// less boilerplate for us, more verbose for user
multi.mustGetProvider('etheruem').getTransactionReceipt('0x...')
```

typescript/optics-provider/src/optics/events.ts
```diff
@@ -0,0 +1,115 @@
...
+import {BigNumber} from "@ethersproject/bignumber";
+import {core} from "@optics-xyz/ts-interface";
+
+export type OpticsStatus = {
+    status: string;
+    events: OpticsEvent[];
+}
+export type OpticsEvent = DispatchEvent | UpdateEvent | ProcessEvent;
+
+interface RichEvent {
```
anna-carroll, line 14: TODO: i wanted to think of some kind of interface that would include all of the information we want for an event which would be returned to a frontend - particularly: 
1. which chain it happened on (so FE can link to block explorer)
2. timestamp (for nice UI things) 
3. transaction hash (again for link to block explorer)

this PR doesn't actually implement all of these things yet. 
need to think about it more in the morning.

typescript/optics-provider/src/optics/events.ts
```diff
@@ -0,0 +1,115 @@
+import {Log} from "@ethersproject/abstract-provider";
+import {OpticsContext} from "./OpticsContext";
+import {BaseContract, Event} from "ethers";
+import {TypedEventFilter} from "@optics-xyz/ts-interface/dist/optics-core/commons";
+import {BigNumber} from "@ethersproject/bignumber";
+import {core} from "@optics-xyz/ts-interface";
+
+export type OpticsStatus = {
+    status: string;
```
anna-carroll, line 9: TODO: make an `enum` for this

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
```diff
@@ -35,75 +23,158 @@ enum MessageStatus {
...
+    }
+    events.push(process);
+    return {
+      status: "Processed", // the message was processed
+      events
+    };
+  }
+
+  async confirmAt(): Promise<BigNumber> {
+    const update = await this.replicaUpdateEvent();
```
anna-carroll, line 166: thought: perhaps we store the event to a field on the class when we query it, and only query it if the field on the class is currently null, so we only query the event once per `OpticsMessage` instance

typescript/optics-provider/src/optics/events.ts
```diff
@@ -0,0 +1,115 @@
+import {Log} from "@ethersproject/abstract-provider";
+import {OpticsContext} from "./OpticsContext";
+import {BaseContract, Event} from "ethers";
+import {TypedEventFilter} from "@optics-xyz/ts-interface/dist/optics-core/commons";
+import {BigNumber} from "@ethersproject/bignumber";
+import {core} from "@optics-xyz/ts-interface";
```
prestwich, line 6: nit: combine the two imports from `ts-interface`

typescript/optics-provider/src/optics/events.ts
```diff
@@ -0,0 +1,115 @@
...
+// PROCESS
+export interface ProcessEvent extends RichEvent {
+    args: {
+        messageHash: string;
+        success: boolean;
+        returnData: string;
+    };
+}
+
+export class MultiEvents {
```
prestwich, line 56: why is this a class? seems to have no state worth speaking of

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
```diff
@@ -35,75 +23,158 @@ enum MessageStatus {
...
+  readonly home: core.Home;
+  readonly replica: core.Replica;
   protected context: OpticsContext;
+  protected eventProvider: MultiEvents;
 
-  constructor(receipt: ContractReceipt, context: OpticsContext) {
+  constructor(event: DispatchEvent, receipt: TransactionReceipt, context: OpticsContext) {
+    this.event = event;
     this.receipt = receipt;
+    this.messageHash = event.args.messageHash;
```
prestwich, line 51: can we do these as `get` functions?

typescript/optics-provider/src/optics/messages/BridgeMessage.ts
```diff
@@ -125,40 +127,70 @@ class BridgeMessage extends OpticsMessage {
...
     }
   }
 
+  static async fromTransactionHash(
+      nameOrDomain: string | number,
+      transactionHash: string,
+      context: OpticsContext,
+  ): Promise<AnyBridgeMessage[]> {
+    const opticsMessages: OpticsMessage[] = await OpticsMessage.fromTransactionHash(nameOrDomain, transactionHash, context);
+    return opticsMessages.map(opticsMessage => BridgeMessage.fromOpticsMessage(opticsMessage, context));
```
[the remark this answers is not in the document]
anna-carroll: > This will error if any message in the tx is not a bridge message, no?

yes you're right. we should return the successfully parsed bridge messages. it seems like perhaps not necessary to do `Promise.allSettled` given that it's synchronous to parse an `OpticsMessage` into a `BridgeMessage`? could do a try/catch or something similar to OpticsMessage.fromReceipt:
```
    let messages: OpticsMessage[] = [];
    for (const log of receipt.logs) {
      const event = MultiEvents.tryDispatchEvent(log);
      if (event) {
        const message = new OpticsMessage(event, receipt, context);
        messages.push(message);
      }
    }
    return messages;
```
^^ would be nice to be able to do this type of functionality in less or cleaner lines of code.. basically `map` but with a try/catch and discarding errors

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
```diff
@@ -35,75 +23,158 @@ enum MessageStatus {
...
   readonly committedRoot: string;
   readonly message: ParsedMessage;
-
+  readonly home: core.Home;
+  readonly replica: core.Replica;
   protected context: OpticsContext;
+  protected eventProvider: MultiEvents;
 
-  constructor(receipt: ContractReceipt, context: OpticsContext) {
+  constructor(event: DispatchEvent, receipt: TransactionReceipt, context: OpticsContext) {
```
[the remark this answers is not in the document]
anna-carroll: yeah, i agree, it seems ugly to me. the point of passing both is that it specifies **which event** to use, because ofc the `receipt` has an array of potentially multiple. another way to do this would be to pass the `receipt` and the `logIndex` of the event in question, but it seemed to me like this would make for weirder code instantiating the class.

the goal is: we want to instantiate the class for a single, specific event - keeping in mind that there may be multiple in one transaction or block; but we also want to preserve all of the context included in the transaction & block

the (ugly) solution i came up with (in an admittedly quick manner) was to pass both specific & general

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
anna-carroll: sure :~)

typescript/optics-provider/src/optics/events.ts
anna-carroll: i don't think you can import them both from the same path because `TypedEventFilter` is not exported from the `index.ts` file

typescript/optics-provider/src/optics/events.ts
anna-carroll: yeah.. it doesn't :/ i think maybe i got annoyed with passing `context` into every function... but these should probably just be functions

typescript/optics-provider/src/optics/messages/BridgeMessage.ts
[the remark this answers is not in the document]
anna-carroll: https://github.com/celo-org/optics-monorepo/pull/842/commits/e28fea8b777fdf84b0c9f49183c6a749c0d13f82

typescript/optics-provider/src/optics/events.ts
anna-carroll: https://github.com/celo-org/optics-monorepo/pull/842/commits/f7d89158a0f91cc53712b890b4256739cbb2145a

typescript/optics-provider/src/optics/events.ts
anna-carroll: https://github.com/celo-org/optics-monorepo/pull/842/commits/d5e3cae00daf9c0805e8af61e4b7284586fe82dc

typescript/optics-provider/src/optics/examples/sendCoins.ts
```diff
@@ -1,6 +1,6 @@
 import * as ethers from 'ethers';
 
-import { mainnet } from '.';
+import { mainnet } from '../index';
```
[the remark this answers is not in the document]
anna-carroll: https://github.com/celo-org/optics-monorepo/pull/842/commits/2955f6d1596750aa8244d1a686fb4c86c44e6676

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
anna-carroll: https://github.com/celo-org/optics-monorepo/pull/842/commits/b324d9ba7f3a38d7723acbdf82139a3b47de9f6c
hmmmm this is basically a shitty version of cacheing.. and significantly less readable.. wdyt @prestwich

ErinHales:
Going to look over this first thing after I get some coffee, but wanted to say thanks for stepping in here. Absolutely incredible 🌸

typescript/optics-provider/src/optics/events.ts
anna-carroll: okay here's what I came up with: https://github.com/celo-org/optics-monorepo/pull/842/commits/f3c973ac1a7a8894d36406d99b9e32f310337c0c

typescript/optics-provider/src/optics/events.ts
```diff
@@ -0,0 +1,110 @@
...
+// PROCESS
+export interface ProcessEvent extends RichEvent {
+    args: {
+        messageHash: string;
+        success: boolean;
+        returnData: string;
+    };
+}
+
+export async function eventToRichEvent(context: OpticsContext, nameOrDomain: string | number, event: Event): Promise<RichEvent> {
```
anna-carroll, line 53: note: this costs 2 extra RPC calls per event. will be expensive / slow for big arrays of events. 

need to do two calls because `timestamp` is not included on `TransactionReceipt`

typescript/optics-provider/src/optics/events.ts
anna-carroll: https://github.com/celo-org/optics-monorepo/pull/842/commits/25129882e7dad895b2255d763f369b9efe78f943

typescript/optics-provider/src/optics/events.ts
prestwich: ugh fair. I guess file an issue to get the `dist` out of our import path at least 🤔

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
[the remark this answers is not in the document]
prestwich: I also appreciate doing it _right_ the first time. This is a fine kludge, but maybe in the future we change to passing the index in the receipt's log array?

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
prestwich: seems reasonable. computing them eagerly in the constructor will also catch any parsing errors

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
prestwich: I think in the future we should consider a log-caching layer on the multi-provider. but lets omit this logic for now.

typescript/optics-provider/src/optics/events.ts
```diff
@@ -0,0 +1,115 @@
...
+    const receipt: TransactionReceipt = await event.getTransactionReceipt();
+    const transaction: TransactionResponse = await event.getTransaction();
+    return {
+        nameOrDomain,
+        ... event,
+        ...receipt,
```
prestwich, line 63: nit, inconsistent whitespace here

prestwich, line 63: The question above about `RichReceipt` is basically "why are we erasing info here by dumpint the receipt into the rich event instead of just attaching the receipt to the rich event?

typescript/optics-provider/src/optics/events.ts
```diff
@@ -0,0 +1,115 @@
...
+    if (POLYGON_IDENTIFIERS.includes(nameOrDomain)) {
+    return getPolygonEvents(context, nameOrDomain, contract, logFilter, startBlock, endBlock);
+    }
+    return contract.queryFilter(logFilter, startBlock, endBlock);
+}
+
+// TODO: have created a major kludge to deal with Polygon RPC errors.. is there a better way?
+const POLYGON_IDENTIFIERS = ["polygon", "matic", 1886350457];
+const POLYGON_FIRST_BLOCK = 18895794;
+const POLYGON_MAX_BLOCKS = 2500;
```
prestwich, line 85: should be `1999`?

typescript/optics-provider/src/optics/events.ts
```diff
@@ -0,0 +1,115 @@
...
+
+export async function getEvents(context: OpticsContext, nameOrDomain: string | number, contract: BaseContract, logFilter: TypedEventFilter<any, any>, startBlock?: number, endBlock?: number): Promise<Array<Event>> {
+    if (POLYGON_IDENTIFIERS.includes(nameOrDomain)) {
+    return getPolygonEvents(context, nameOrDomain, contract, logFilter, startBlock, endBlock);
+    }
+    return contract.queryFilter(logFilter, startBlock, endBlock);
+}
+
+// TODO: have created a major kludge to deal with Polygon RPC errors.. is there a better way?
+const POLYGON_IDENTIFIERS = ["polygon", "matic", 1886350457];
```
prestwich, line 83: module-level consts should be at top of file

typescript/optics-provider/src/optics/events.ts
```diff
@@ -0,0 +1,115 @@
...
+    return {
+        nameOrDomain,
+        ... event,
+        ...receipt,
+        timestamp: transaction.timestamp!, // timestamp will exist in TransactionResponse because transaction has definitely mined (because TransactionReceipt already returned)
+        allTransactionLogs: receipt.logs,
+    };
+}
+
+export async function getRichEvents(context: OpticsContext, nameOrDomain: string | number, contract: BaseContract, logFilter: TypedEventFilter<any, any>, startBlock?: number, endBlock?: number): Promise<Array<RichEvent>> {
```
prestwich, line 69: should these be module-level functions, seems like they ought to be methods on the `MultiProvider` or `OpticsContext`?

typescript/optics-provider/src/optics/events.ts
```diff
@@ -0,0 +1,115 @@
...
+    }
+    return contract.queryFilter(logFilter, startBlock, endBlock);
+}
+
+// TODO: have created a major kludge to deal with Polygon RPC errors.. is there a better way?
+const POLYGON_IDENTIFIERS = ["polygon", "matic", 1886350457];
+const POLYGON_FIRST_BLOCK = 18895794;
+const POLYGON_MAX_BLOCKS = 2500;
+
+export async function getPolygonEvents(context: OpticsContext, nameOrDomain: string | number, contract: BaseContract, logFilter: TypedEventFilter<any, any>, startBlock?: number, endBlock?: number): Promise<Array<Event>> {
```
prestwich, line 87: proposed _additional_ kludge:
allow user to specify an early exit after finding a single event. E.g.  a `singleton: bool = false` argument. this allows us to prevent crawling the entire range if we're looking for a particular event

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
```diff
@@ -27,84 +15,225 @@ export type ParsedMessage = {
...
+export type OpticsStatus = {
+  status: MessageStatus;
+  events: OpticsEvent[];
+}
+
+export enum MessageStatus {
+  Dispatched = 0,
+  Included = 1,
+  Relayed = 2,
+  Processed = 3
```
prestwich, line 27: inconsistent enum declaration. either use explicit `=` on all elements of both enums or on just the first element of both

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
```diff
@@ -27,84 +15,225 @@ export type ParsedMessage = {
...
-  readonly destinationAndNonce: BigNumber;
-  readonly committedRoot: string;
+  readonly dispatchEvent: DispatchEvent;
+  readonly receipt: TransactionReceipt;
   readonly message: ParsedMessage;
-
+  readonly home: core.Home;
+  readonly replica: core.Replica;
   protected context: OpticsContext;
+  protected storedHomeUpdateEvent: UpdateEvent | undefined;
```
prestwich, line 54: I want to drop this caching but, in the future, use a `type EventCache = { ... } ` for things like this

prestwich: `protected cache = { ... }`

prestwich: that way it's fewer lines of declaration (more legible) and you don't have to make wordy variable names to indicate intent

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
```diff
@@ -27,84 +15,225 @@ export type ParsedMessage = {
...
+      transactionHash: string,
+      context: OpticsContext,
+  ): Promise<OpticsMessage> {
+    const provider = context.mustGetProvider(nameOrDomain);
+    const receipt = await provider.getTransactionReceipt(transactionHash);
+    return OpticsMessage.singleFromReceipt(receipt!, context);
+  }
 
-    this.context = context;
+  async homeUpdateEvent(): Promise<UpdateEvent | undefined> {
```
prestwich, line 117: need to note in docs that this function may produce false positives, if the updater has signed an update on the committedRoot that does not include this message. these should be uncommon in practice. The only way to know _for sure_ is to keep an indexed set of updates and leaves :(

prestwich, line 117: are these `_____UpdateEvent` methods intended to be public API? I think not, right? if not, consider making them private.

Also, not a fan of the naming. no need to name it `____Event` when we return an event. the type signature makes that clear. What about `homeUpdate`

typescript/optics-provider/src/optics/events.ts
anna-carroll: nevermind, we can't have nice things - `timestamp` not available on `TransactionReceipt` actually, and `getBlock` throws for Celo

anna-carroll: https://github.com/celo-org/optics-monorepo/issues/848

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
anna-carroll: I had this initially, the thing is, that makes one of the functions `process() `which is confusing naming IMO, sounds like it's doing something on-chain

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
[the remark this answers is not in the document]
prestwich: resolved in latest with introduction of `Annotated<T>`

prestwich:
@anna-carroll please put me out of my misery and/or review this PR

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
prestwich: I settled on `get____` without event. it's a little unclear for process but seems fine. 

Are they intended to be public api tho?

anna-carroll: ah shoot, you are right.

anna-carroll: i think it's good for them to be public API, if someone wants to query for just one

yourbuddyconner (changes requested):
I want to use this. 

Please add .env to .gitignore.

prestwich:
-_-

typescript/optics-provider/src/optics/messages/OpticsMessage.ts
anna-carroll: ugh.

typescript/optics-provider/src/optics/events/fetch.ts
```diff
@@ -0,0 +1,146 @@
...
+export function annotate<U extends Result, T extends TypedEvent<U>>(
+  domain: number,
+  receipt: TransactionReceipt,
+  event: T,
+): Annotated<T> {
+  return {
+    domain,
+    receipt,
+    name: event.eventSignature?.split('(')[0],
+    event,
```
anna-carroll, line 29: i wonder if it wouldn't be beneficial to have the `annotatedEvent` have all the same fields as a regular event (e.g. do `...event` here) so that people can interact with it as if its a normal event (with a few extra fields)

typescript/optics-provider/src/optics/events/fetch.ts
```diff
@@ -0,0 +1,146 @@
...
+  // or current block number
+  let lastBlock;
+  if (!endBlock) {
+    const provider = context.mustGetProvider(domain.id);
+    lastBlock = await provider.getBlockNumber();
+  } else {
+    lastBlock = endBlock;
+  }
+  // query domain pagination limit at a time, concurrently
+  const callArgs = [];
```
anna-carroll, line 112: this loop isn't used. i'll open a PR

anna-carroll (commented):
looking beautiful 😭  thank you @prestwich

typescript/optics-provider/src/optics/events/fetch.ts
anna-carroll: https://github.com/celo-org/optics-monorepo/pull/858

PR #843, opened 2021-09-24
fix: fix overloaded index config var

yourbuddyconner:


emberian:
this can't be set with an env var due to the underscore. rename to indexon or something?

emberian (approved):

PR #844, opened 2021-09-24
fix: reduce page size from 9999 to 2000

yourbuddyconner:


emberian:
does this slide in under some api limits?

emberian (approved):
hope these work!

yourbuddyconner:
Our API limit is 10k blocks but when we query that many with all the agents at once the requests begin to timeout (likely due to overloading the backend).

PR #850, opened 2021-09-28
fix: fix processor index/nonce conflation and add per-replica nonce storage

### LINKED ISSUE #851: bug: Replica processors share the same key for storing per-replica nonces

ltchang2019:
Bug: multiple `Replica` processors were sharing the same key for storing their latest inspected nonces
Changes:
-  key latest nonces by replica domain
- improve docs for pub db store functions
- Processor `IntGaugeVec` tracks Replica processor next_nonces
- standardize on `leaf` and remove all refs to `leaf_hash`

Closes #851

### DIFF
--- a/rust/agents/processor/src/processor.rs
+++ b/rust/agents/processor/src/processor.rs
@@ -76,7 +76,7 @@ impl Replica {
                     .unwrap_or_default();
 
                 self.next_nonce
-                    .with_label_values(&[self.replica.name(), AGENT_NAME])
+                    .with_label_values(&[self.home.name(), self.replica.name(), AGENT_NAME])
                     .set(next_nonce as i64);
 
                 info!(
--- a/rust/agents/processor/src/processor.rs
+++ b/rust/agents/processor/src/processor.rs
@@ -27,7 +27,6 @@ use optics_core::{
 
 use crate::{prover_sync::ProverSync, settings::ProcessorSettings as Settings};
 
-const LAST_INSPECTED: &str = "lastInspected";
 const AGENT_NAME: &str = "processor";
 
 /// The replica processor is responsible for polling messages and waiting until they validate
@@ -72,7 +71,7 @@ impl Replica {
                 // 5. Submit the proof to the replica
                 let mut next_message_index: u32 = self
                     .home_db
-                    .retrieve_decodable("", LAST_INSPECTED)?
+                    .retrieve_latest_nonce(domain)?
                     .map(|n: u32| n + 1)
                     .unwrap_or_default();
 
@@ -106,11 +105,8 @@ impl Replica {
                         .await
                     {
                         Ok(true) => {
-                            self.home_db.store_encodable(
-                                "",
-                                LAST_INSPECTED,
-                                &next_message_index,
-                            )?;
+                            self.home_db
+                                .store_latest_nonce(domain, next_message_index)?;
                             next_message_index += 1;
                         }
                         Ok(false) => {
--- a/rust/agents/processor/src/processor.rs
+++ b/rust/agents/processor/src/processor.rs
@@ -39,7 +39,7 @@ pub(crate) struct Replica {
     home_db: HomeDB,
     allowed: Option<Arc<HashSet<H256>>>,
     denied: Option<Arc<HashSet<H256>>>,
-    next_nonce: prometheus::IntGaugeVec,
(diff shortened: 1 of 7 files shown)


rust/agents/processor/src/processor.rs
```diff
@@ -267,7 +262,7 @@ impl Processor {
             .new_int_gauge(
                 "next_message_index",
                 "Index of the next message to inspect",
-                &["replica", "agent"],
+                &["home", "agent"],
```
ltchang2019, line 265: I only understand these prometheus objects at a very surface level but this `IntGaugeVec` doesn't look like it's getting used anywhere. @prestwich did we just forget to actually track/update this one?

rust/optics-core/src/db/home_db.rs
```diff
@@ -235,6 +263,9 @@ impl HomeDB {
     }
 
     /// Store a proof by its leaf index
+    ///
+    /// Keys --> Values:
+    /// - `leaf_index` --> `proof`
     pub fn store_proof(&self, leaf_index: u32, proof: &Proof) -> Result<(), DbError> {
         debug!(leaf_index, "storing proof in DB");
         self.store_keyed_encodable(PROOF, &leaf_index, proof)
```
ltchang2019, line 271: I'm confused, are these _all caps_ prefixes supposed to describe the key or the value? It looks like most cases describe the key but its inconsistent everywhere... should consolidate to standard

emberian (dismissed):

rust/agents/processor/src/processor.rs
prestwich: so the `&[&str]` here is identifiers for groups of "labels" on the data. It should probably be `&["home", "replica", "agent"]` for this group of metrics. and then when we make a specific metric

Each metric is updated on L78 in the processor 

```
                self.next_nonce
                    .with_label_values(&[self.replica.name(), AGENT_NAME])
                    .set(next_nonce as i64);
```

rust/optics-core/src/db/home_db.rs
prestwich: I think they should describe the value. Feel free to change to make consistent

rust/agents/processor/src/processor.rs
prestwich: if updating the label array, also update the `with_label_values` invocation

rust/agents/processor/src/processor.rs
```diff
@@ -314,20 +310,26 @@ impl OpticsAgent for Processor {
...
         let denied = self.denied.clone();
 
-        let next_message_index = self.next_message_index.clone();
-
         tokio::spawn(async move {
             let replica = replica_opt.ok_or_else(|| eyre!("No replica named {}", name))?;
             let home_name = home.name().to_owned();
 
+            let next_nonce = metrics
+                .new_int_gauge(
```
prestwich, line 318: instead of making a new one, put the one on the `Processor` into an arc (and rename it appropriately), and give the arc to each replica processor struct. then have each replica processor set given its labels.

So no metric for `next_message_index` and a shared `Arc<IntGaugeVec>` for `next_nonce`

rust/optics-core/src/db/typed_db.rs
```diff
@@ -28,6 +28,7 @@ impl TypedDB {
     fn full_prefix(&self, prefix: impl AsRef<[u8]>) -> Vec<u8> {
         let mut full_prefix = vec![];
         full_prefix.extend(self.type_prefix.as_ref() as &[u8]);
+        full_prefix.extend("_".as_bytes());
```
prestwich, line 31: this will cause a complete DB resync right? cc @yourbuddyconner

ltchang2019: Yes this will require us to scrap the existing db

prestwich (changes requested):
metrics update

rust/optics-core/src/db/typed_db.rs
ltchang2019: Either way the key prefix changes will make us do this

prestwich: yeah, just need to give conner a headsup when we do this. it's still relatively cheap, so no real downside but better to know than not know

prestwich (approved):

PR #854, opened 2021-09-29
fix: wait for token approval in send

prestwich:


anna-carroll (approved):
🥜

typescript/optics-provider/src/optics/OpticsContext.ts
```diff
@@ -269,7 +269,8 @@ export class OpticsContext extends MultiProvider {
     const approved = await fromToken.allowance(senderAddress, bridgeAddress);
     // Approve if necessary
     if (approved.lt(amount)) {
-      await fromToken.approve(bridgeAddress, amount, overrides);
+      let tx = await fromToken.approve(bridgeAddress, amount, overrides);
```
anna-carroll, line 272: spicy idea.. what if we construct these txs with subsequent nonces and `await Promise.all` their execution?

prestwich: the metamask interface will not estimate gas properly for the second one, and will flag it to the user as a potential error

PR #859, opened 2021-09-30
fix: fix kathy sync bug related to `ethers-rs` concurrency

### LINKED ISSUE #857: bug: kathy nonce collision
I was debugging with @ltchang2019 and believe I have a hypothesis for why Kathy is crash-looping. 

When booting, kathy spawns N tasks, each responsible for dispatching messages to a particular configured replica. Issues arise when each task tries to dispatch a message at the same time or close together, because the individual tasks attempt to use the same account Nonce. 

I believe this might necessitate a fix at the ethers-rs level. 

CC @prestwich for thoughts.

ltchang2019:
- add lock to kathy and have her acquire and hold for duration of `home.dispatch(...).await` calls to ensure only one dispatch call happening at one time (temp fix to ether-rs concurrency bug)
- lock does not wrap home because that would require changing `AgentCore` setup

Closes #857

### DIFF
--- a/rust/agents/kathy/src/kathy.rs
+++ b/rust/agents/kathy/src/kathy.rs
@@ -88,10 +88,9 @@ impl OpticsAgent for Kathy {
                             recipient = message.recipient
                         );
 
-                        let _guard = home_lock.lock().await;
+                        let guard = home_lock.lock().await;
                         home.dispatch(&message).await?;
-
-                        // Drop home_lock after dispatch
+                        drop(guard);
                     }
                     _ => {
                         info!("Reached the end of the static message queue. Shutting down.");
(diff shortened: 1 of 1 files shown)


rust/agents/kathy/src/kathy.rs
```diff
@@ -86,13 +87,19 @@ impl OpticsAgent for Kathy {
                             destination = message.destination,
                             recipient = message.recipient
                         );
+
+                        let _guard = home_lock.lock().await;
                         home.dispatch(&message).await?;
+
+                        // Drop home_lock after dispatch
```
prestwich, line 94: drop `_guard`

prestwich (changes requested):
comment nit

prestwich (approved):

PR #863, opened 2021-10-01
feature: `HomeIndexer` stores update metadata

### LINKED ISSUE #822: DB: have prover sync store information about update contents
Currently prover sync traverses from update to update, ingesting leaves

-  prover_sync should store metadata about tree structure, 
    - which leaves an update contains (from_index, to_index)
- homeindexer should store metadata about update events
    - block height, timestamp

any other good things to store @yourbuddyconner ?

ltchang2019:
- home indexer stores `UpdateMeta` struct in db (keyed by `update.new_root`)
- `UpdateMeta` currently only has `block_number` but we can add any additional fields if needed

Closes #822

### DIFF
--- a/rust/optics-core/src/db/home_db.rs
+++ b/rust/optics-core/src/db/home_db.rs
@@ -1,4 +1,5 @@
 use crate::db::{DbError, TypedDB, DB};
+use crate::UpdateMeta;
 use crate::{
     accumulator::merkle::Proof, traits::RawCommittedMessage, utils, Decode, Encode, OpticsMessage,
     SignedUpdate,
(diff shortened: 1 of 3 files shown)


rust/optics-core/src/db/home_db.rs
```diff
@@ -19,6 +20,7 @@ static PREV_ROOT: &str = "update_prev_root_";
 static PROOF: &str = "proof_";
 static MESSAGE: &str = "message_";
 static UPDATE: &str = "update_";
+static BLOCK_NUMBER: &str = "block_number_";
```
prestwich, line 23: nit: should this be `update_meta_` or something similar?

prestwich (approved):

PR #881, opened 2021-10-08
refactor: simplify `proversync` and remove incremental merkle usage

luketchang:
- the reason why deleting persistence and restarting fixes the stuck processor is that `from_disk` checks for missing proofs and fills them all
- tried to find bug in `proversync::spawn` code that leads to missing db proofs and it's likely a range error or something nuanced
- may be missing something but usage of the incremental merkle seems to way overcomplicate the `proversync` and it doesn't seem to provide any benefit right now
- it serves as a check before committing to the in-memory prover tree but no matter what happens, if the prover root becomes mismatched with what we expect from signed updates, we will bail (no way to handle gracefully anyways so may as well just commit to the in-memory prover first)


ltchang2019:
Will merge #879 first then move base to `main` and merge this one

PR #914, opened 2021-10-26
feature: request retrier middleware

yourbuddyconner:
Related to #916

### DIFF
--- a/rust/chains/optics-ethereum/src/retrying.rs
+++ b/rust/chains/optics-ethereum/src/retrying.rs
@@ -9,6 +9,7 @@ use serde::{de::DeserializeOwned, Serialize};
 use thiserror::Error;
 use tokio::time::sleep;
 
+/// A naive exponential backoff provider
 #[derive(Debug, Clone)]
 pub struct RetryingProvider {
     inner: Http,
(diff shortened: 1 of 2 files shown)


rust/chains/optics-ethereum/src/retrying.rs
```diff
@@ -0,0 +1,81 @@
...
+};
+use serde::{de::DeserializeOwned, Serialize};
+use thiserror::Error;
+use tokio::time::sleep;
+
+/// A naive exponential backoff provider
+#[derive(Debug, Clone)]
+pub struct RetryingProvider {
+    inner: Http,
+    max_requests: usize,
```
emberian, line 16: there is a setter, but no getter. awkward!

```suggestion
    pub max_requests: usize,
```

will let anyone use it or change it if they need.

rust/chains/optics-ethereum/src/retrying.rs
```diff
@@ -0,0 +1,81 @@
...
+    where
+        T: Debug + Serialize + Send + Sync,
+        R: Serialize + DeserializeOwned,
+    {
+        for i in 0..self.max_requests {
+            {
+                let res = self.inner.request(method, &params).await;
+                if res.is_ok() {
+                    return Ok(res.expect("checked"));
+                }
```
emberian, line 66:
```suggestion
                if let Ok(d) = self.inner.request(method, &params).await {
                  return Ok(d)
                }
```

emberian (approved):
some style nits, but this is g2g

PR #925, opened 2021-11-02
refactor/feature: split up indexing/contract/db logic + make watcher functional again

### LINKED ISSUE #870: Move db logic from `optics-ethereum` to `optics-core`
- The chain-specific home currently keeps a DB and puts things into it.
- instead, this should be handled 1 layer up in `optics-core`. `Homes` enum 
- chain should be unaware of the DB

```
pub struct CachingHome {
    db: HomeDb
    home: Homes
}

// replace `impl Home for Homes`
impl Home for CachingHome {
    // for each function, check the DB first (if appropriate)
    // cache afterwards (if appropriate)
    // basically replicate the existing functionality in `EthereumHome`
}

```

ltchang2019:
`CommonIndexer/HomeIndexer`: Interface by which chain-specific homes/replicas bubble up event data (each chain will have a different implementation/library used retrieve this data which is why we need a trait to provide common interface). Bubbles up data to a ContractSync, which takes that data and stores it in agent's db.
`ContractSync`: Drives syncing between agent db and on-chain data in loop. Continuously calls out to an indexer to retrieve event data and stores it in agent's db.
`CachingHome/CachingReplica`: Both contain a `sync` function which spawns a new `ContractSync` to keep db up-to-date.

Closes #870


rust/config/development/watcher-partial.json
```diff
@@ -4,30 +4,26 @@
     "key": "",
     "type": "hexKey"
   },
-  "connectionManagers": [
-    {
-      "address": "0x0000000000000000000000000000000000000000",
-      "domain": "60",
-      "name": "ethereum",
+  "managers": {
+    "kovanbridgerouter": {
```
prestwich, line 8: `camelCase` for json?

rust/optics-core/src/traits/mod.rs
```diff
@@ -102,6 +105,19 @@ pub trait Common: Sync + Send + std::fmt::Debug {
...
+    /// Submit a double update for slashing
+    async fn double_update(
+        &self,
+        double: &DoubleUpdate,
+    ) -> Result<TxOutcome, ChainCommunicationError>;
+}
+
+/// Interface for retrieving event data emitted by both the home and replica
+#[async_trait]
+pub trait CommonEvents: Common + Send + Sync + std::fmt::Debug {
```
prestwich, line 120: can you provide rationale for splitting the trait?

rust/optics-base/src/home.rs
```diff
@@ -1,14 +1,204 @@
...
+}
+
+impl CachingHome {
+    /// Instantiate new CachingHome
+    pub fn new(home: Arc<Homes>, db: OpticsDB, indexer: Arc<Indexers>) -> Self {
+        Self { home, db, indexer }
+    }
+
+    /// Return handle on home object
+    pub fn home(&self) -> Arc<Homes> {
```
prestwich, line 34: We should consider moving the `Homes` arc internally, so that Homes is always clonable. the same pattern we did with OpticsDB

prestwich (changes requested):
Okay the main thing I think this needs is documentation of the layout of all these types. How do all the traits relate to eachother and why?

rust/optics-core/src/traits/mod.rs
ltchang2019: Wasn't strictly necessary but thinking to split because `Homes` switches on chain specific contract calls while retrieval of event data is chain agnostic (just db fetches). Other alternative was to scrap `Homes` entirely and to just put all methods (contract calls and event data fetches) in `CachingHome` and have `CachingHome` implement `Home` with both contract calls and event data fetches.

ltchang2019:
Documentation:
- Indexer traits: https://github.com/celo-org/optics-monorepo/pull/925/files#diff-6bd64866866ea21992b5ba0925c8dcaa2efca43e5d5409d369e4bec27f85be4bR1-R8
- ContractSync struct: https://github.com/celo-org/optics-monorepo/pull/925/files#diff-8915f2b78fa646fab4b83ec99793b39e5c908ed3cb74af8dc49f403ea5d40b94R16-R20
- HomeEvents/CommonEvents: `Homes` and `Replicas` abstractions, which switches over chain-specific contract methods are retained and implement `Home` and Replicas`. `CachingHome` and `CachingReplica` also provide event data (updates, messages) and thus implement `HomeEvents` and `CommonEvents`

prestwich (approved):
approved, but needs to run in dev before deployment

PR #932, opened 2021-11-28
Optics v2 deployment configs and outputs

timmoreton:
* Fix deploy to enable EIP1559 gas pricing
* Fix configured recovery mode timelock not getting set in deployment
* Add avalanche
* Move governance to Celo
* Set new multisigs for recovery manager and governance
* Set optimisticSeconds to 30mins


nambrot:
V2 has been running for, so I think we can merge this

nambrot (approved):

PR #934, opened 2021-12-11
Optics v2 new deployment for optics-provider

nambrot:
This PR updates the optics-provider package to include v2 deployments for both staging and production


nambrot:
Going to merge in since it was mostly @tkporter

PR #942, opened 2022-01-10
Update contracts-metrics for productionCommunity

nambrot:
- Bumps the provider package to paginate Avalanche RPC event lookups
- Updates the `contracts-metrics` package to observe the v2 deployment

### DIFF
--- a/typescript/contract-metrics/src/config.ts
+++ b/typescript/contract-metrics/src/config.ts
@@ -9,11 +9,11 @@ const environment = process.env.ENVIRONMENT ?? 'development'
 let networks = []
 switch (environment) {
   case 'production':
-    networks = ['ethereum', 'celo', 'polygon']
+    networks = ['celo', 'ethereum', 'polygon', 'avalanche']
     break;
 
   default:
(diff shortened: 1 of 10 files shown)


typescript/contract-metrics/src/config.ts
```diff
@@ -9,11 +9,11 @@ const environment = process.env.ENVIRONMENT ?? 'development'
 let networks = []
 switch (environment) {
   case 'production':
-    networks = ['ethereum', 'celo', 'polygon']
+    networks = ['celo', 'ethereum', 'polygon', 'avalanche']
     break;
 
   default:
-    networks = ['alfajores', 'rinkeby', 'kovan']
+    networks = ['alfajores', 'kovan']
```
asaj, line 16: Why did this change?

asaj (dismissed):

typescript/contract-metrics/src/config.ts
nambrot: Not sure, readded it since it doesn't seem to actually break anything like I thought

asaj (approved):

PR #977, opened 2022-12-07
Switch packages to use the correct npm org

dckesler:
Even those these packages aren't published we're switching them to use the correct npm organization in case someone were to actually publish something malicious to the npm registry using these names.


pputman-clabs (dismissed):
approving to fix the namespace on the npm packages.

pputman-clabs (approved):
approving to fix npm registry namespace for security issues.

